PR #74, opened 2022-06-01
feat: conform to new subgraph structure

FiboApe:


src/graphql/getLeaderboardPosition.graphql.ts
```diff
@@ -44,44 +43,79 @@ const getLeaderboardPositions = gql`
...
         withdrawn
-        idleSwapped
-        permissions {
-          id
-          operator
-          permissions
-        }
+        toWithdraw
       }
       history: actionsHistory(orderBy: createdAtTimestamp, orderDirection: asc) {
```
0xged, line 59: `actionsHistory` will not longer exist

src/graphql/getPosition.graphql.ts
```diff
@@ -45,44 +44,79 @@ const getPosition = gql`
...
         withdrawn
-        idleSwapped
-        permissions {
-          id
-          operator
-          permissions
-        }
+        toWithdraw
       }
       history: actionsHistory(orderBy: createdAtTimestamp, orderDirection: asc, first: $first, skip: $skip) {
```
0xged, line 60: Same :)

0xged (commented):
A few comments: 
- `nextSwapAvailableAt` is getting removed, I don't see that reflected here.
- `totalDeposits` also is being renamed to `totalDeposited`, pelase change 🙏 

Other than those two things, everything looks good!

PR #80, opened 2022-06-10
Add profit loss graph

FiboApe:


nchamo (approved):
Looks good, we just need to make a change before we merge it

PR #95, opened 2022-07-19
fix: use real balance when modifying positions

FiboApe:


nchamo (approved):
Thank u for fixing this so quickly!

PR #117, opened 2022-09-13
fix: add v char before version

feraway:
`TRANSACTION_TYPES.NEW_POSITION` was missing a `"v"` char in the id, so then on `src/services/positionService.ts:160` we failed to get the position and update it with data from the positions api call, hence creating a new position that appeared as duplicate until we refreshed the client.


FiboApe (approved):
🤦 The amount of times I've had this error

0xged:
Congrats on your first PR @feraway! 🎉

PR #120, opened 2022-09-20
fix check by using lowecase

feraway:
The check `position.user === walletService.getAccount()` failed because `walletService.getAccount()` had uppercases while `position.user` didn't, so we were never shown the "Approve" button in the position edition modal.


FiboApe (approved):
Nice Catch!

PR #126, opened 2022-09-23
MF-307: feat(meanapiservice): add 30% more to gasLimit on calls to companion

FiboApe:


nchamo (approved):
Looks great!

src/services/meanApiService.ts
```diff
@@ -35,6 +36,17 @@ export default class MeanApiService {
     this.client = client;
   }
 
+  async addGasLimit(tx: TransactionRequest): Promise<TransactionRequest> {
+    const singer = this.walletService.getSigner();
```
feraway, line 40: `"singer" !== "signer"` :p

feraway (approved):

src/services/meanApiService.ts
FiboApe: Good catch! Will fix

PR #127, opened 2022-09-23
MF-312: feat(positiondetails): add breakdown toggle to position details

FiboApe:


feraway (approved):
LGTM! Toggle looks nice and the "detailed" view is not verbose at all, very succinct and to the point.

PR #134, opened 2022-10-02
Avoid unnecesary renders caused by fetching new data

feraway:
Avoid "Confirm Operation / Allow" button flickr on position acceptance by correcting useEffect dependencies, memoization and default values

### DIFF
--- a/src/hooks/useAllowance.ts
+++ b/src/hooks/useAllowance.ts
@@ -8,40 +8,44 @@ import { useBlockNumber } from 'state/block-number/hooks';
 import useCurrentNetwork from './useCurrentNetwork';
 import useWalletService from './useWalletService';
 
-function useAllowance(
-  from: Token | undefined | null
-): [{ token: Token; allowance: string } | undefined, boolean, string?] {
-  const [isLoading, setIsLoading] = React.useState(false);
+type Allowance = {
+  token: Token;
+  allowance: string | undefined;
+};
+
+type AllowanceResponse = [Allowance, boolean, string?];
+
+const dummyToken: Allowance = { token: EMPTY_TOKEN, allowance: undefined };
+
+function useAllowance(from: Token | undefined | null): AllowanceResponse {
   const walletService = useWalletService();
-  const [result, setResult] = React.useState<{ token: Token; allowance: string } | undefined>(undefined);
-  const [error, setError] = React.useState<string | undefined>(undefined);
+  const [state, setState] = React.useState({ isLoading: false, result: dummyToken, error: undefined });
   const hasPendingTransactions = useHasPendingTransactions();
   const prevFrom = usePrevious(from);
   const prevPendingTrans = usePrevious(hasPendingTransactions);
-  const account = usePrevious(walletService.getAccount());
+  const prevAccount = usePrevious(walletService.getAccount());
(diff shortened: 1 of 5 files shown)


src/hooks/useAllowance.ts
```diff
@@ -8,70 +8,76 @@ import { useBlockNumber } from 'state/block-number/hooks';
...
     async function callPromise() {
       if (from) {
         try {
           const promiseResult = await walletService.getAllowance(from);
-          setResult(promiseResult);
-          setError(undefined);
+          setState((prevState) => ({ ...prevState, result: promiseResult, error: undefined, isLoading: false }));
         } catch (e) {
-          setError(e);
+          /* eslint-disable  @typescript-eslint/no-unsafe-assignment, @typescript-eslint/no-unsafe-member-access */
```
FiboApe, line 40: We can set e as string to avoid this lint warning.

Btw instead of enabling and disabling eslint-rules we can just do
```
// eslint-disable-next-line <eslint rule>
```

And it just works for one line

src/hooks/useAllowance.ts
```diff
@@ -8,70 +8,76 @@ import { useBlockNumber } from 'state/block-number/hooks';
...
+
+type AllowanceResponse = [Allowance, boolean, string?];
+
+const dummyToken: Allowance = { token: EMPTY_TOKEN, allowance: undefined };
+
+function useAllowance(from: Token | undefined | null): AllowanceResponse {
   const walletService = useWalletService();
-  const [result, setResult] = React.useState<{ token: Token; allowance: string } | undefined>(undefined);
-  const [error, setError] = React.useState<string | undefined>(undefined);
+  const [state, setState] = React.useState({ isLoading: false, result: dummyToken, error: undefined });
```
FiboApe, line 22: could we use object destructuring here?

src/hooks/useBalance.ts
```diff
@@ -9,67 +9,69 @@ import useCurrentNetwork from './useCurrentNetwork';
 import useWalletService from './useWalletService';
 
 function useBalance(from: Token | undefined | null): [BigNumber | undefined, boolean, string?] {
-  const [isLoading, setIsLoading] = React.useState(false);
   const walletService = useWalletService();
-  const [result, setResult] = React.useState<BigNumber | undefined>(undefined);
-  const [error, setError] = React.useState<string | undefined>(undefined);
+  const [state, setState] = React.useState<{ isLoading: boolean; result?: BigNumber; error?: string }>({
```
FiboApe, line 13: Same here, lets use object des-tructuring

src/hooks/useBalance.ts
```diff
@@ -9,67 +9,69 @@ import useCurrentNetwork from './useCurrentNetwork';
...
     async function callPromise() {
       if (from) {
         try {
           const promiseResult = await walletService.getBalance(from.address);
-          setResult(promiseResult);
-          setError(undefined);
+          setState({ isLoading: false, result: promiseResult, error: undefined });
         } catch (e) {
-          setError(e);
+          setState((prevState) => ({ ...prevState, error: e, isLoading: false }));
```
FiboApe, line 36: we probably don't want to spread the previous state, since we want to set result as undefined if it has failed

PR #139, opened 2022-10-21
Allow user to choose min allowance with split button

feraway:
- Allow user to choose min allowance with split button
- Expand split button to resemble full width block buttons
- Change primary color for it to be compliant with contrast color codes

image


src/common/modal/index.tsx
```diff
@@ -86,6 +87,7 @@ const Modal: React.FC<ModalProps> = ({
   showCloseButton,
   actions,
   children,
+  Component,
```
FiboApe, line 90: I'm 100% not sold on having a "Component" prop.

2 Alternatives here:
- We can change the actions definition to accept a `split` property where that defines wether we show a split button or a normal button.
- We change the actions definition to either accept the object it accepts right now, Buttons and SplitButtons, this would end up having a type of
```
interface ActionsObject {
    label: React.ReactNode;
    onClick: () => void;
    disabled?: boolean;
    color?: keyof typeof ButtonTypes;
    variant?: 'text' | 'outlined' | 'contained';
  }[];
  
  type actions = (ActionsObject | Button | SplitButton)[]
  ```
  
  Then when we render it we just check for the type and render the appropiate child, if its an ActionObject it renders it like we have it right now, and if its a button or SplitButton the map just returns that

File: src/common/modify-settings-modal/index.tsx [hunk 2]
```diff
@@ -322,7 +323,13 @@ const ModifySettingsModal = ({ position, open, onCancel }: ModifySettingsModalPr
           </Typography>
         ),
       });
-      const result = await walletService.approveToken(fromToUse, fromHasYield, version);
+
+      const result = await walletService.approveToken(
+        fromToUse,
+        fromHasYield,
+        version,
+        amount ? parseUnits(amount, from.decimals) : undefined
```
FiboApe, line 331: This is not correct, they only need to approve the difference between what is already on the position and what they actually want the position to have in the end

`remainingLiquidity.sub(BigNumber.from(frequencyValue).mul(parseUnits(rate, fromToUse.decimals))).abs(),`

src/common/modify-settings-modal/index.tsx
```diff
@@ -358,24 +365,42 @@ const ModifySettingsModal = ({ position, open, onCancel }: ModifySettingsModalPr
...
+  let Component;
+
   if (needsToApprove) {
-    actions = [
-      {
-        color: 'primary',
-        variant: 'contained',
-        label: (
+    Component = (
+      <SplitButton
```
FiboApe, line 372: Can we also add a help tooltip icon to the main button like we had on the button before it?

The copy would be 
```
You must give the Mean Finance smart contracts permission to use your {token}
```

src/common/modify-settings-modal/index.tsx
```diff
@@ -358,24 +365,42 @@ const ModifySettingsModal = ({ position, open, onCancel }: ModifySettingsModalPr
...
+        onClick={() => handleApproveToken()}
+        text={
           <FormattedMessage
             description="Allow us to use your coin"
-            defaultMessage="Approve {token}"
+            defaultMessage="Approve Max {token}"
             values={{
-              token: fromToUse.symbol,
+              token: (from && from.symbol) || '',
```
FiboApe, line 379: this should be `fromToUse` since it will always show as ETH for protocol tokens for example

src/common/modify-settings-modal/index.tsx
```diff
@@ -358,24 +365,42 @@ const ModifySettingsModal = ({ position, open, onCancel }: ModifySettingsModalPr
...
+        disabled={!!hasPendingApproval}
+        variant="contained"
+        color="primary"
+        options={[
+          {
+            text: (
+              <FormattedMessage
+                description="Allow us to use your coin"
+                defaultMessage="Approve {fromValue} {token}"
+                values={{ token: (from && from.symbol) || '', fromValue }}
```
FiboApe, line 392: Same as before for fromTouse

src/common/token-picker/index.tsx
```diff
@@ -182,7 +182,7 @@ const Row = ({ index, style, data: { onClick, tokenList, tokenKeys, yieldOptions
         </StyledTokenTextContainer>
         {!!availableYieldOptions.length && (
           <StyledTokenTextContainer>
-            <Typography variant="caption" component="span" color="#2CC941">
+            <Typography variant="caption" component="span" color="#007E36">
```
FiboApe, line 185: lets not change this green

src/home/swap-container/components/swap/index.tsx
```diff
@@ -253,7 +254,12 @@ const Swap = ({
           </Typography>
         ),
       });
-      const result = await walletService.approveToken(from, !!(shouldEnableYield && fromYield?.tokenAddress));
+      const result = await walletService.approveToken(
```
FiboApe, line 257: Could we also have a new TRANSACTION_TYPES that is 'APPROVE_TOKEN_EXACT' and we send both the token and the amount approved, so then we can later show both on the snackbar 
"Approve X <TOKEN> to be used" and on the wallet transaction timeline the same text

src/common/modify-settings-modal/index.tsx
```diff
@@ -364,16 +384,41 @@ const ModifySettingsModal = ({ position, open, onCancel }: ModifySettingsModalPr
...
+          <>
+            <FormattedMessage
+              description="Allow us to use your coin"
+              defaultMessage="Approve Max {token}"
+              values={{
+                token: fromToUse.symbol,
+              }}
+            />
+            <StyledTooltip
+              title={`You must give the Mean Finance smart contracts permission to use your ${fromToUse.symbol}`}
```
FiboApe, line 396: Lets set a FormattedMessage for the title, so we can later translate it

src/common/modify-settings-modal/index.tsx
```diff
@@ -364,16 +384,41 @@ const ModifySettingsModal = ({ position, open, onCancel }: ModifySettingsModalPr
...
+          </>
         ),
-        onClick: handleApproveToken,
+        onClick: () => handleApproveToken(),
         disabled: !!hasPendingApproval,
+        options: [
+          {
+            text: (
+              <FormattedMessage
+                description="Allow us to use your coin"
```
FiboApe, line 410: this description should be different since they are used as the uniq ids for the messages once we want tot ranslate them

src/common/modify-settings-modal/index.tsx
```diff
@@ -364,16 +384,41 @@ const ModifySettingsModal = ({ position, open, onCancel }: ModifySettingsModalPr
...
+              title={`You must give the Mean Finance smart contracts permission to use your ${fromToUse.symbol}`}
+              arrow
+              placement="top"
+            >
+              <HelpOutlineIcon fontSize="small" />
+            </StyledTooltip>
+          </>
         ),
-        onClick: handleApproveToken,
+        onClick: () => handleApproveToken(),
```
FiboApe, line 404: no need for this right? should just be onClick: handleApproveToken

src/home/swap-container/components/swap/index.tsx
```diff
@@ -618,19 +626,44 @@ const Swap = ({
...
+          <>
+            <FormattedMessage
+              description="Allow us to use your coin"
+              defaultMessage="Approve Max {token}"
+              values={{
+                token: (from && from.symbol) || '',
+              }}
+            />
+            <StyledTooltip
+              title={`You must give the Mean Finance smart contracts permission to use your ${from?.symbol || ''}`}
```
FiboApe, line 638: Same as before.

Maybe we should move all of this into it's own component?

FiboApe (commented):
Mostly good left some comments.


Nice work here!

src/common/allowance-split-button/index.tsx
```diff
@@ -0,0 +1,64 @@
...
+      arrow
+      placement="top"
+    >
+      <HelpOutlineIcon fontSize="small" />
+    </StyledTooltip>
+  );
+};
+
+const AllowanceSplitButton = (props: AllowanceSplitButtonProps) => {
+  const { onClick, options, text, disabled, symbol } = props;
```
FiboApe, line 43: I would actually pass the token and the amount here, let the button define which text is going to show.

I would only put this props for the button

```
disabled?: boolean
onMaxApprove: () => void
onApproveExact: (amount: BigNumber) => void
token: Token
amount: BigNumber
```

FiboApe (approved):
Nice work here!

PR #150, opened 2022-11-04
feat(*): add smol domains integration

FiboApe:


src/services/walletService.ts
```diff
@@ -90,6 +90,21 @@ export default class WalletService {
       return ens;
     }
 
+    const currentNetwork = await this.getNetwork();
+
+    if (currentNetwork.chainId === NETWORKS.arbitrum.chainId) {
```
0xged, line 95: Seems weird to check if its arbitrum and also have a mapping for arbitrum chain id? Maybe you can check if the smolDomainInstance its undefined, then check ENS or viceversa? Just to avoid this hardcode?

0xged (commented):
Smol *hehe* comment. I think in the future we might end up generalizing this by having some interface like `NameSupportService`, and check for ENS, SmolDomains, UnstoppableDomains, and so on.

src/services/walletService.ts
FiboApe: Smol domain can exist in other chains and for now I think we only want to integrate them for arbitrum, this handles the case if in the future we add the other chains where smol domains operates

PR #160, opened 2022-11-14
feat(aggregator): add quote data information

FiboApe:


src/aggregator/swap-container/components/quote-data/index.tsx
```diff
@@ -0,0 +1,77 @@
...
+            values={{ network: networkName }}
+          />
+        </Typography>
+        <Typography variant="body2">
+          {quote?.gas.estimatedCostInUSD ? `$${quote.gas.estimatedCostInUSD}` : '-'}
+        </Typography>
+      </StyledQuoteDataItem>
+      <StyledQuoteDataItem>
+        <Typography variant="body2" color="inherit">
+          <FormattedMessage description="quoteDataRate" defaultMessage="Minimum recieved:" />
```
feraway, line 65: Please correct to "received"

feraway (approved):
LGTM! Great work :D 

The only suggestion I have is to add a tooltip on the "Minimum received", to let the user understand the source of that value, since the values on QuoteData and on SwapQuotes might be different

image

PR #177, opened 2022-11-22
feat(token-inputs): add usd values to show on inputs

FiboApe:


src/aggregator/swap-container/components/step1/index.tsx
```diff
@@ -139,13 +170,24 @@ const SwapFirstStep = React.forwardRef<HTMLDivElement, SwapFirstStepProps>((prop
...
                 onChange={handleFromValueChange}
-                withBalance
-                balance={balance}
                 token={from}
-                withMax
                 fullWidth
+                usdValue={parseFloat(fromPrice || '0').toFixed(2)}
               />
-              <TokenButton token={from} onClick={() => startSelectingCoin(from || emptyTokenWithAddress('from'))} />
+              <StyledTokenButtonContainer>
```
feraway, line 177: I don't think it's apparent to the user that this is a clickable button

feraway (approved):

PR #185, opened 2022-11-24
Use different token pickers for aggregator and dca

FiboApe:


src/common/modal/index.tsx
```diff
@@ -43,7 +43,7 @@ const StyledDialogActions = styled(DialogActions)`
 `;
 
 const StyledDialog = styled(Dialog)`
-  display: flex;
+  // display: flex;
```
feraway, line 46: removing `display: flex` looks fishy as there are others flex selectors in place

FiboApe: Other ones implement their own

feraway (approved):
LGTM :)

PR #190, opened 2022-11-24
feat(aggregator): add tooltip to minimum recieved

FiboApe:


feraway (approved):
Looks great :) I think it's clearer now

PR #191, opened 2022-11-25
fix: avoid all failure if one subgraph is down

0xsambugs:


0xsambugs:
Closing in favor of #192

PR #199, opened 2022-11-30
Sort swap routes in ui

feraway:
Default quote fetching to "most-profit", save it in state and only sort by "least-gas" when needed. Because the array of DEX options is small, I avoided using a memo since I don't see a perf gain.


src/hooks/useSwapOptions.ts
```diff
@@ -65,11 +63,12 @@ function useSwapOptions(
               debouncedTo,
               debouncedIsBuyOrder ? undefined : parseUnits(debouncedValue, debouncedFrom.decimals),
               debouncedIsBuyOrder ? parseUnits(debouncedValue, debouncedTo.decimals) : undefined,
-              debouncedSorting,
+              'most-profit',
```
FiboApe, line 66: we should probably drop this into a constant in `src/constants/aggregator.ts` file

src/hooks/useSwapOptions.ts
```diff
@@ -138,7 +133,13 @@ function useSwapOptions(
     return [undefined, false, undefined, fetchOptions];
   }
 
-  return [result || prevResult, isLoading, error, fetchOptions];
+  let resultToReturn = result || prevResult;
+
+  if (sorting === 'least-gas' && resultToReturn) {
+    resultToReturn = [...resultToReturn].sort((a, b) => (a.gas.estimatedCost.lt(b.gas.estimatedCost) ? -1 : 1));
```
FiboApe, line 139: We should probably React.useMemo this, so it does not recalculate it on every re-render

src/hooks/useSwapOptions.ts
feraway: Sounds good!

src/hooks/useSwapOptions.ts
feraway: I don't think that is an issue, the list of routes is small and the check is very simple. Imo the load of useMemo would be worse than the check itself.

src/hooks/useSwapOptions.ts
feraway: Fixed

FiboApe (approved):

PR #201, opened 2022-12-01
Add aggregator aggregator

FiboApe:


feraway (approved):
Looks great! :D Just a couple of minor comments regarding console.logs and css selectors. The Meta Aggregator is close to starting it's reign 🦾

PR #204, opened 2022-12-02
feat(aggregator): do not refresh routes while user is swapping

FiboApe:


feraway (approved):

feraway:
Code looks good and the feature is working as expected :)

Suggestion: what do you think about avoiding the refresh as soon as the user presses swap? So when the modal appears before the user confirms in metamask?

Also related: what do you think about disabling the Swap button while quotes are being fetched?

PR #208, opened 2022-12-05
feat(*): add yield blacklist for creation of positions

FiboApe:


src/services/yieldService.ts
```diff
@@ -27,7 +27,7 @@ export default class YieldService {
     this.client = client;
   }
 
-  async getYieldOptions(chainId?: number): Promise<YieldOptions> {
+  async getYieldOptions(chainId?: number, ignoreBlacklist = false): Promise<YieldOptions> {
```
nchamo, line 30: Can we rename`ignoreBlacklist` to `useBlacklist` or `listenToBlacklist`? (Here and in all the other files)

nchamo (approved):
Looks good 👍

PR #211, opened 2022-12-06
Revert do not use state

FiboApe:


nchamo (approved):

src/home/swap-container/components/swap/index.tsx
```diff
@@ -195,18 +196,14 @@ const Swap = ({
...
-      parseUnits(fromValue, from.decimals).gt(BigNumber.from(0)) &&
-      frequencyValue &&
-      BigNumber.from(frequencyValue).gt(BigNumber.from(0)) &&
-      from &&
-      formatUnits(parseUnits(fromValue, from.decimals).div(BigNumber.from(frequencyValue)), from.decimals)) ||
-    '0';
-
-  const rateUsdPrice = parseUsdPrice(from, (rate !== '' && parseUnits(rate, from?.decimals)) || null, usdPrice);
+  try {
+    rateForUsdPrice = (rate !== '' && parseUnits(rate, from?.decimals)) || null;
```
feraway, line 202: Why would this fail?

src/home/swap-container/components/swap/index.tsx
```diff
@@ -217,6 +214,19 @@ const Swap = ({
   const shouldEnableYield =
     yieldEnabled && (fromCanHaveYield || toCanHaveYield) && isAtLeastAWeek && hasEnoughUsdForYield;
 
+  React.useEffect(() => {
+    if (!from) return;
+    setRate(
```
feraway, line 219: I think some of this checks will set rate to false or undefined when they fail, and we are only checking if `rate !== ''` in line 202. if this is so, we might send an unsupported value to parseUnits

src/home/swap-container/components/swap/index.tsx
```diff
@@ -408,11 +419,21 @@ const Swap = ({
     if (!from) return;
     setModeType(FULL_DEPOSIT_TYPE);
     setFromValue(newFromValue);
+    setRate(
+      (newFromValue &&
+        parseUnits(newFromValue, from.decimals).gt(BigNumber.from(0)) &&
```
feraway, line 424: Same concern as before here

feraway (changes requested):
Left some concerns in the comments, we might not need any changes but setting the PR to "Request changes" just in case.

feraway (approved):

feraway:
Approved after talking it over with @FiboApe

PR #218, opened 2022-12-07
feat(*): add next swap available at on positions

FiboApe:


feraway (approved):
LGTM :D great addition

PR #221, opened 2022-12-22
Aggregator Feedback

FiboApe:


src/common/transaction-confirmation/index.tsx
```diff
@@ -130,12 +130,14 @@ const TransactionConfirmation = ({ shouldShow, handleClose, transaction }: Trans
...
       setSuccess(true);
+      // eslint-disable-next-line @typescript-eslint/no-floating-promises
       confetti({
         particleCount: 100,
         spread: 70,
         angle: 60,
         origin: { x: 0 },
       });
+      // eslint-disable-next-line @typescript-eslint/no-floating-promises
       confetti({
```
feraway, line 142: 🎊

src/state/token-lists/reducer.ts
```diff
@@ -63,15 +76,27 @@ export const getDefaultByUrl = () => ({
...
   activeAggregatorLists: [
     'https://raw.githubusercontent.com/Mean-Finance/token-list/main/mean-finance.tokenlist.json',
     'tokens.1inch.eth',
+    'custom-tokens',
   ],
   byUrl: getDefaultByUrl(),
+  customTokens: {
+    name: 'custom-tokens',
+    logoURI: '',
+    timestamp: new Date().getTime(),
```
feraway, line 93: What is this timestamp for?

src/common/modal/index.tsx
```diff
@@ -111,6 +113,7 @@ const Modal: React.FC<ModalProps> = ({
       classes={classes}
       onClose={handleClose}
       PaperProps={fullHeightProps}
+      keepMounted={keepMounted}
```
feraway, line 116: I don't see a reference to `keepMounted` in `StyledDialog`

src/services/walletService.ts
```diff
@@ -277,6 +277,10 @@ export default class WalletService {
       return Promise.resolve({ token, allowance: formatUnits(MaxUint256, token.decimals) });
     }
 
+    if (addressToCheck === NULL_ADDRESS) {
```
feraway, line 280: What's the reason behind this check?

feraway (approved):

src/common/transaction-confirmation/index.tsx
FiboApe: 🎊

src/state/token-lists/reducer.ts
FiboApe: it's the timestamp of the last updated value, it is from the schema from how you build official Token Lists

src/services/walletService.ts
FiboApe: For wrapping/unwrapping ETH/WETH 0x and other dexes will return the null address as the allowanceTarget since for wrapping/unwrapping you need to call the WETH contract directly and does not need any approval

PR #284, opened 2023-02-14
refactor: simplify sdk service

0xsambugs:
We are now simplifying the sdk service by using the sdk directly

### DIFF
--- a/package.json
+++ b/package.json
@@ -109,7 +109,7 @@
     "@mean-finance/dca-v2-core": "^3.0.0",
     "@mean-finance/dca-v2-periphery": "^3.0.0",
     "@mean-finance/oracles": "^2.0.0",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -109,7 +109,7 @@
     "@mean-finance/dca-v2-core": "^3.0.0",
     "@mean-finance/dca-v2-periphery": "^3.0.0",
     "@mean-finance/oracles": "^2.0.0",
-    "@mean-finance/sdk": "0.0.22",
+    "@mean-finance/sdk": "^0.0.26",
```
FiboApe, line 112: I would like to keep this hardfixed since patch versions on `0.0.x` in standard versioning can mean breaking changes

FiboApe (approved):

PR #290, opened 2023-02-15
feat(aggregator): allow automatically changing networks for wallets t…

FiboApe:
…hat support it


feraway (approved):
Works like a charm! The only small issue I found is that while automatically changing wallets, there is a short flickr where the "action" button of /create says "Change network to..." and is enabled. As a suggestion, perhaps we could disabled it and change the message to "Changing network to..."

PR #300, opened 2023-02-16
refactor: remove dist from sdk import path

0xged:
this will make it easier to link the SDK. blz do test it locally without all the links. remember to `yarn install --force`


FiboApe:
Are you sure this works? Lint does not seem to recognize the imports

PR #316, opened 2023-03-03
feat(aggregator): disable inputs while loading

FiboApe:


feraway:
For some reason the MAX button still triggers even if disabled 🤔 

https://user-images.githubusercontent.com/113060391/223626199-0ae558c9-620c-4ef9-a1b7-aecc2fcbbc63.mov

PR #351, opened 2023-03-22
feat(permissions): allow permissioned users to interact with positions

feraway:
- Allow permissioned users to interact with positions
- Modify positions permission setter to simplify adding permissions to the Mean Finance Companion
- Disable confirmation button on modify-settings-modal when no change is made
- Allow WITHDRAW (REDUCE) with wrapped token from split button in position details


src/common/modify-settings-modal/index.tsx
```diff
@@ -649,6 +682,13 @@ const ModifySettingsModal = ({ position, open, onCancel }: ModifySettingsModalPr
                 />
               </FormGroup>
             )}
+            {defaultWrapped && (
+              <FormattedMessage
+                description="useWrappedTokenHint"
+                defaultMessage="You can only use {token} because the Mean Finance Compainion doesn't have the required permissions"
```
FiboApe, line 688: Typo with companion here

src/common/modify-settings-modal/index.tsx
```diff
@@ -123,7 +139,29 @@ const ModifySettingsModal = ({ position, open, onCancel }: ModifySettingsModalPr
...
+  const hasNotChangedPosition = remainingLiquidity
+    .sub(parseUnits(fromValue || '0', fromToUse.decimals))
+    .eq(BigNumber.from(0));
+
+  const companionHasRequiredPermission = isIncreasingPosition
+    ? !!companionPermissions.INCREASE
+    : !!companionPermissions.REDUCE;
+
+  const defaultWrapped =
+    !!shouldShowWrappedProtocolSwitch && !accountPermissions.isOwner && !companionHasRequiredPermission;
```
FiboApe, line 158: no need for double negation here

src/common/terminate-modal/index.tsx
```diff
@@ -92,9 +107,9 @@ const TerminateModal = ({
       const hasPermission = await positionService.companionHasPermission(position, PERMISSIONS.TERMINATE);
       if (hasWrappedOrProtocol && protocolBalance.gt(BigNumber.from(0))) {
         if (hasProtocolToken) {
-          terminateWithUnwrap = !useProtocolToken;
+          terminateWithUnwrap = !useWrappedProtocolToken;
         } else {
-          terminateWithUnwrap = useProtocolToken;
+          terminateWithUnwrap = useWrappedProtocolToken;
```
FiboApe, line 112: have you actually tied this? bc now the conditions should be reversed right?

src/common/terminate-modal/index.tsx
```diff
@@ -237,6 +253,14 @@ const TerminateModal = ({
             />
           </FormGroup>
         )}
+        {defaultWrapped && (
+          <Typography variant="body1">
+            <FormattedMessage
+              description="terminate warning wrapped companion"
+              defaultMessage="The Mean Finance Companion doesn't have the required permissons, you can only terminate with a Wrapped Protocol Token."
```
FiboApe, line 260: Can we add the token symbol here?

src/home/positions-list/components/position-controls/index.tsx
```diff
@@ -171,6 +179,26 @@ const PositionControls = ({
...
+  );
+
+  const accountPermissions = createPermissionsObject(
+    position?.permissions?.find((permission) => permission.operator === account.toLowerCase())?.permissions,
+    account.toLowerCase() === position.user.toLowerCase()
+  ) as PositionPermissions;
+
+  const mergedPermissions =
+    hasYield || isToProtocolToken
+      ? mergeCompanionPermissions(accountPermissions, companionPermissions)
```
FiboApe, line 200: I'm not sure why we need to merge companion permissions based on some parameters

src/position-detail/position-summary-controls/index.tsx
```diff
@@ -107,18 +139,9 @@ const PositionSummaryControls = ({
...
-    position.to.address === PROTOCOL_TOKEN_ADDRESS;
-  const shouldDisableArrow =
-    isPending ||
-    disabled ||
-    (!shouldShowWithdrawWrappedToken && BigNumber.from(position.remainingLiquidity).lte(BigNumber.from(0)));
-
   return (
     <PositionControlsContainer>
-      {showExtendedFunctions && (
+      {showExtendedFunctions && (mergedPermissions.INCREASE || mergedPermissions.REDUCE) && (
```
FiboApe, line 144: This condition should probably be part of showExtendedFunctions?

src/position-detail/position-summary-controls/index.tsx
```diff
@@ -201,12 +203,54 @@ const PositionSummaryControls = ({
...
+                    disabled:
+                      isPending || disabled || BigNumber.from(position.remainingLiquidity).lte(BigNumber.from(0)),
+                    onClick: () => onWithdrawFunds(false),
+                  },
+                ]
+              : []),
           ]}
         />
       )}
+      {shouldDisableArrow && mergedPermissions.WITHDRAW && !mergedPermissions.REDUCE && (
```
FiboApe, line 231: Why is reduce taken into account here?

feraway: If the user has permissions to REDUCE we need the split button because they can withdraw base funds

src/services/positionService.ts
```diff
@@ -248,14 +272,21 @@ export default class PositionService {
                   (position.pair.swaps[0] && parseInt(position.pair.swaps[0].executedAtTimestamp, 10)) ||
                   position.createdAtTimestamp,
                 pairNextSwapAvailableAt: position.createdAtTimestamp.toString(),
+                ...(!!position.permissions && { permissions: position.permissions }),
```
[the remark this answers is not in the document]
feraway: It's not necessary to use a ternary operator since boolean values spread to nothing, example: 

image

However, if you prefer using a ternary operator because you believe it's clearer, I will include your suggestion and procede that way from now on forward whenever I implement a short circuit of object spreading.

src/state/position-permissions/hooks.ts
```diff
@@ -1,11 +1,42 @@
...
+  WITHDRAW?: string;
+  TERMINATE?: string;
+};
+
+export const createPermissionsObject = (permissions: Permission[] = [], isOwner = false) => {
+  const permissionsToUse = isOwner ? ['WITHDRAW', 'TERMINATE', 'REDUCE', 'INCREASE'] : permissions;
+  return permissionsToUse.reduce((acc, curr) => ({ ...acc, [curr]: true }), { isOwner });
+};
+
+export function mergeCompanionPermissions(
```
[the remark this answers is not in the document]
feraway: The goal is getting a permission object that only has ```true``` for permissions that are both present in the ```accountsPermissions``` and the ```companionPermissions``` so for example, if you send it
```
// companion
{ isOwner: false, TERMINATE: true, WITHDRAW: true}

// account

{ isOwner: false, TERMINATE: true, REDUCE: true}
```

You will receive

```
{ isOwner: false, TERMINATE: true }
```

Which is the only overlap in truthy permissions. This merging of permissions is used in cases where there is yield of protocol tokens, to reduce the amount of inline conditionals ```hasYield ? accountPermissions.TERMINATE && companionPermissions.TERMINATE : accountPermission.TERMINATE &&...``` . So I conditionally merged user and companion permissions in some components, based on yield or protocol token (like in ```src/home/positions-list/components/position-controls/index.tsx:198``` ) to avoid redundant conditional code. WDYT?

If the user is owner it just returns their permissions (which should aways be all of them) because the companion is not needed.

src/home/positions/index.tsx
```diff
@@ -154,15 +190,59 @@ const Positions = () => {
...
+                  </Typography>
+                }
+              />
+            )}
           </StyledTabs>
         </StyledTabsContainers>
         <StyledPositionsContainer>
           <StyledPaper variant="outlined">
-            {tabIndex === 0 ? <CurrentPositions isLoading={!hasLoadedPositions || isLoading} /> : <History />}
+            {(() => {
```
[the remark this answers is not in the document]
feraway: I think that an inline switch statement adds clarity to the tabs, so I decided to use an inline IIFE instead of inline jsx like
```
{tabIndex === 0 && <PositionList....>}
{tabIndex === 1 && <PositionList...}
```

but its just a syntax preference so I'm ok refactoring to the above example if you prefer, or however you believe it's clearer.

src/services/meanApiService.ts
```diff
@@ -168,7 +168,6 @@ export default class MeanApiService {
 
   async terminateUsingOtherTokens(
     id: string,
-    recipientUnswapped: string,
```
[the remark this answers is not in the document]
feraway: I'm not clear on this comment. I just realized that recipientUnswapped was being passed to the fn but it wasn't being used in the fn itself, so I just removed it from the arguments of the functions and the parameters in the places the function is called.

src/services/positionService.ts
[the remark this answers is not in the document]
FiboApe: nah, lets keep it as you made it

src/graphql/getPermissionedPositions.graphql.ts
```diff
@@ -0,0 +1,72 @@
...
+
+const getPermissionedPositions = gql`
+  query getPermissionedPositions($address: String!) {
+    positions(where: { permissions_: { operator: $address } }) {
+      id
+      createdAtTimestamp
+      totalDeposited
+      totalSwaps
+      totalSwapped
+      totalWithdrawn
```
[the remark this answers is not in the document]
feraway: All the `total....` fields are referenced in the code

PR #357, opened 2023-03-28
refactor: update sdk and api url

0xsambugs:
In here we are updating to the latest version of the SDK and we are also using the new API endpoint


src/services/sdkService.ts
```diff
@@ -37,17 +37,17 @@ export default class SdkService {
...
                 list: {
                   type: 'api',
-                  baseUri: ({ chainId }: { chainId: number }) =>
-                    `https://api.mean.finance/v1/swap/networks/${chainId}/quotes`,
+                  baseUri: ({ chainId, sourceId }: { chainId: number; sourceId: SourceId }) =>
+                    `https://api.mean.finance/v1/swap/networks/${chainId}/quotes/${sourceId}`,
                   sources: SOURCES_METADATA,
                 },
-                sourceIds: ['uniswap', 'odos', 'rango', '0x', 'firebird'],
+                sourceIds: ['uniswap', 'odos', 'rango', '0x', 'firebird', 'changelly'],
```
0xsambugs, line 50: We are adding changelly!

src/services/sdkService.ts
```diff
@@ -28,7 +28,7 @@ export default class SdkService {
     this.sdk = buildSDK({
       provider: {
         source: {
-          type: 'only-first-provider-that-supports-chain',
+          type: 'prioritized',
```
0xsambugs, line 31: Just renamed

src/services/sdkService.ts
```diff
@@ -37,17 +37,17 @@ export default class SdkService {
           type: 'overridable-source-list',
           lists: {
             default: {
-              type: 'default',
+              type: 'local',
```
0xsambugs, line 40: Another rename

src/services/sdkService.ts
```diff
@@ -206,9 +205,13 @@ export default class SdkService {
       return this.walletService.getCustomToken(address);
     }
 
-    const tokenResponse = await this.sdk.tokenService.getTokens({ addresses: [{ addresses: [address], chainId }] });
+    const tokenResponse = await this.sdk.metadataService.getMetadataForChain({
```
0xsambugs, line 208: We can now fetch the name!! 🎉 

I marked it as required, is that ok? Other options would be "best effort" or "can ignore"

FiboApe (approved):

PR #359, opened 2023-03-30
feat(*): add jbrl dashboard

FiboApe:


src/jbrl-competition/frame/index.tsx
```diff
@@ -0,0 +1,326 @@
...
+  );
+
+  const onCreatePosition = () => {
+    dispatch(changeMainTab(0));
+    history.push('/create/137/0xf2f77fe7b8e66571e0fca7104c4d670bf1c8d722/');
+  };
+
+  const [yieldOptions, isLoadingYieldOptions] = useYieldOptions(NETWORKS.polygon.chainId);
+
+  const positions = React.useMemo(() => data?.positions || [], [data]);
```
feraway, line 69: Why are you memoizing a field reference?

feraway (approved):

src/jbrl-competition/frame/index.tsx
FiboApe: So it doesn't change on every re-render due to reference checking

PR #370, opened 2023-04-06
fix(sdkservice): do not fetch custom token when address is not valid

FiboApe:


feraway (commented):
The regex works, but the problem I found is that when you enter an invalid address, the screen just appears as an infinite loading with the Skeleton component from MUI never leaving. I would consider displaying a message saying that the address is invalid.

PR #374, opened 2023-04-08
feat: add mean source

0xsambugs:
We are now updating the SDK, which basically adds the new Mean Finance source 🔥

### DIFF
--- a/src/services/sdkService.ts
+++ b/src/services/sdkService.ts
@@ -7,9 +7,9 @@ import { BigNumber } from 'ethers';
 import { SwapOption, Token } from 'types';
 import { AxiosInstance } from 'axios';
 import { toToken } from 'utils/currency';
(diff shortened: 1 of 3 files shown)


src/services/sdkService.ts
```diff
@@ -98,8 +99,8 @@ export default class SdkService {
         ...(gasSpeed ? { gasSpeed: { speed: gasSpeed, requirement: 'best effort' } } : {}),
         ...(skipValidation ? { skipValidation } : {}),
         ...(isBuyOrder ? { estimateBuyOrdersWithSellOnlySources: true } : {}),
-        quoteTimeout: '5s',
```
0xsambugs, line 101: Moved the timeout out of the request and onto the config, to make it consistent with all other services

FiboApe (approved):

PR #378, opened 2023-04-13
fix: wido token list

0xged:


0xged:
image
i've also tried this one.

PR #395, opened 2023-05-08
refactor(aggregator): split aggregator components into smaller ones

FiboApe:


feraway (approved):
Great refactor @FiboApe! Aggregator components are now where they belong :D I also really like the change you made making TransactionTypes an enum. LGTM!

PR #414, opened 2023-06-06
refactor: starts using mixpanel proxy

0xged:


.github/workflows/deployToStaging.yml
```diff
@@ -0,0 +1,35 @@
...
+        run: yarn build
+      - name: Configure AWS credentials
+        uses: aws-actions/configure-aws-credentials@v1
+        with:
+          mask-aws-account-id: true
+          aws-access-key-id: ${{ secrets.AWS_ACCESS_KEY_ID }}
+          aws-secret-access-key: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+          aws-region: us-east-1
+      - name: Deploy
+        run: aws s3 sync ./dist s3://${{ secrets.AWS_S3_STAGING }} --exclude "*" --include "bundle.js.map" --include "bundle.js"
```
FiboApe, line 33: dont forget to add this secrets

src/services/eventService.ts
```diff
@@ -1,41 +1,47 @@
...
+  mixpanel: Mixpanel;
 
-  constructor(meanApiService: MeanApiService, providerService: ProviderService) {
-    this.meanApiService = meanApiService;
+  constructor(providerService: ProviderService) {
     this.providerService = providerService;
-    this.sessionId = uuidv4();
+    this.mixpanel = MixpanelLibray.init(
+      '7f62e8855fcbcd72de460ba40b6fc06b',
+      { api_host: 'https://mean-finance-mixpanel-proxy.herokuapp.com' },
```
FiboApe, line 16: This should also go into an env file/secret

src/services/eventService.ts
```diff
@@ -1,41 +1,47 @@
...
+      ' '
+    );
+    this.mixpanel.set_config({ persistence: 'localStorage', ignore_dnt: true });
   }
 
   async getIdentifier() {
     const account = await this.providerService.getAddress();
-
-    return (account && md5(account)) || md5(this.sessionId);
+    return md5(account);
```
FiboApe, line 24: account can be undefined before user connects its wallet, we should maintain the sessionId to at least track users that don't connect their wallet, we could send both so we also track the session

src/services/eventService.ts
```diff
@@ -1,41 +1,47 @@
...
-      'test'
-    );
+    const foundNetwork = find(NETWORKS, { chainId: network.chainId });
+    return new Promise<void>((resolve, reject) => {
+      this.mixpanel.track(
+        action,
+        {
+          chainId: network.chainId,
+          chainName: foundNetwork?.name,
+          ...(extraData || {}),
```
FiboApe, line 36: should we not send the identifier?

0xged: Mixpanel creates it's own identifier based on the device, so it's not really needed. More info [here](https://docs.mixpanel.com/docs/tracking/how-tos/identifying-users).

src/services/eventService.ts
0xged: AFAIK mixpanel already handles this without the need of a `session_id`.  Identifier should be per/user - but sadly mixpanel only allows one user identified per "device" (I can clarify this on a meet).

0xged: Just found out [about this](https://docs.mixpanel.com/docs/analysis/advanced/sessions), we can create a `session_id` but we should def. not use the wallet address for that, just the uuidv4 (will update code).

src/services/eventService.ts
FiboApe: yeah but we created one based on the users wallet, so we can track specific users, do you not still want that?

src/services/eventService.ts
FiboApe: sure, but do you not want to track specific wallets?

src/services/eventService.ts
0xged: I'd rather have 1 user = 1 device than 1 user = 1 wallet. at least 1 user = 1 device will not introduce different type of conversions, etc. I'm researching about Groups on Mixpanel, but don't think it will be useful, sadly.

src/services/eventService.ts
0xged: Nain

FiboApe (approved):

PR #442, opened 2023-09-13
feat(positionService): add timeout to legacy subgraphs

th0rOdinson:
- With this change:
   - All promises, except the last version, will timeout after 30 seconds.
   - If a promise times out or encounters an error, it will resolve with { data: null, error: true }.

### DIFF
--- a/src/services/positionService.ts
+++ b/src/services/positionService.ts
@@ -45,7 +45,7 @@ import { emptyTokenWithAddress } from '@common/utils/currency';
 import { getDisplayToken, sortTokens } from '@common/utils/parsing';
 import gqlFetchAll, { GraphqlResults } from '@common/utils/gqlFetchAll';
 import { doesCompanionNeedIncreaseOrReducePermission } from '@common/utils/companion';
-import { AddFunds, DCAPermission } from '@mean-finance/sdk';
+import { AddFunds, DCAPermission, timeoutPromise } from '@mean-finance/sdk';
 import GraphqlService from './graphql';
 import ContractService from './contractService';
 import WalletService from './walletService';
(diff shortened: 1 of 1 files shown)


src/services/positionService.ts
```diff
@@ -163,8 +163,16 @@ export default class PositionService {
         );
       })
     );
-
-    const results = await Promise.all(promises.map((promise) => promise.catch(() => ({ data: null, error: true }))));
+    const results = await Promise.all(
+      promises.map(async (promise, index) => {
+        const isLastVersion = networksAndVersions[index].version === POSITIONS_VERSIONS[POSITIONS_VERSIONS.length - 1];
```
FiboApe, line 168: There is a constant that you can use `LATEST_VERSION`

src/services/positionService.ts
```diff
@@ -163,8 +163,16 @@ export default class PositionService {
...
+        try {
+          return await timeoutPromise(promise, !isLastVersion ? '30s' : undefined);
+        } catch {
+          return { data: null, error: true };
```
FiboApe, line 172: should we return a promise too? `Promise.resolve({ data: null, error: true })`

src/services/positionService.ts
th0rOdinson: Awesome! Thanks

src/services/positionService.ts
th0rOdinson: I researched the benefits of adding `Promise.resolve()`, and in this case, using `async` ensures the function always returns a promise. However, I can add it for the sake of code consistency.

FiboApe: Nah, its gucci as it is!

FiboApe (approved):

PR #444, opened 2023-09-18
fix(aggregator): Loosing tokens after network switch

th0rOdinson:
## Context
Url was being reseted to `mean.finance/swap/chainId` before triggering a `window.location.reload()`, so selected tokens were lost.
The `chainId` param is already being updated in the URL by the network selector component.

## Demo
1. Enter to the Aggregator page and connect with a wallet that is not a `CHAIN_CHANGING_WALLETS_WITHOUT_REFRESH`
2. Change network using the Network Selector & select tokens to perform a swap
3. Click on the button 'Change network to XXXXX'

Site should refresh and selected tokens should persists.


FiboApe (approved):
Great Work!

PR #447, opened 2023-09-20
feat(aggregator): start tracking failed quotes

th0rOdinson:
From now on, we will be tracking failed quotes in MixPanel, including the error type and the `sourceTimeout` selected by the user. 

### Error Types
_Below are the different types of errors and an example of each as received by the application._

- Timeout: These errors occur when a request times out.

  > Request to https://api.mean.finance/v1/swap/networks/42220/quotes/0x timeouted at 3.5s

- ReferralCode: Errors related to invalid or unregistered referral codes.

  > Example: Odos: failed to calculate a quote between 0xeeeeee…ail\\":\\"Invalid or unregistered referral code\\"}"

- BigIntConversion: Errors arising from attempts to convert an undefined value to a BigInt.

  > Example: Cannot convert undefined to a BigInt

- NetworkRequest: General network request failures.

  > Example: Network request failed

### DIFF
--- a/src/services/aggregatorService.ts
+++ b/src/services/aggregatorService.ts
@@ -11,8 +11,8 @@ import { Interface } from '@ethersproject/abi';
 import ERC20ABI from '@abis/erc20.json';
 import WRAPPEDABI from '@abis/weth.json';
 import { getProtocolToken } from '@common/mocks/tokens';
-import { quoteResponseToSwapOption } from '@common/utils/quotes';
-import { QuoteResponse } from '@mean-finance/sdk/dist/services/quotes/types';
+import { categorizeError, quoteResponseToSwapOption } from '@common/utils/quotes';
+import { FailedQuote, QuoteResponse } from '@mean-finance/sdk/dist/services/quotes/types';
 import { GasKeys, SORT_MOST_PROFIT, SwapSortOptions, TimeoutKey } from '@constants/aggregator';
 import GraphqlService from './graphql';
 import ContractService from './contractService';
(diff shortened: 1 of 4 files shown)


src/services/aggregatorService.ts
```diff
@@ -145,14 +150,32 @@ export default class AggregatorService {
       sourceTimeout
     );
 
-    const filteredOptions = swapOptionsResponse.filter((option) => !('failed' in option)) as QuoteResponse[];
+    // eslint-disable-next-line @typescript-eslint/no-unnecessary-type-assertion
+    const validOptions = (swapOptionsResponse as (QuoteResponse | FailedQuote)[]).reduce(
```
FiboApe, line 154: there is no need for parsing the type, it should already come as `(QuoteResponse | FailedQuote)[]` which could also remove the eslint-disable :)

FiboApe: btw instead of using a reduce, we could use a map, it returns null when the quote is failed, and returns the quote when it is not failed, and then you can use the `compact` function from lodash to filter out the nulls

th0rOdinson: Sure! It's done now. I kept the `as QuoteResponse[]` that was already there, becuase it's not inferring return type from sdkService.getSwapOptions

src/types/aggregator.ts
```diff
@@ -60,3 +60,5 @@ export interface SwapOptionWithFailure extends SwapOption {
 export interface SwapOptionWithTx extends SwapOption {
   tx: QuoteTransaction;
 }
+
+export type QuoteErrorType = 'Timeout' | 'ReferralCode' | 'BigIntConversion' | 'NetworkRequest' | 'Unknown';
```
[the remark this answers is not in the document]
th0rOdinson: Addressed!

FiboApe (approved):
LFG!!!

PR #450, opened 2023-09-25
fix(meanApiService): avoid 0 calls to /to-underlying

th0rOdinson:


src/services/meanApiService.ts
```diff
@@ -62,20 +62,35 @@ export default class MeanApiService {
...
       }
     );
 
-    return underlyingResponse.data.underlying;
+    const finalResponse = underlyingResponse.data.underlying;
+
+    tokensWithoutAmount.forEach((tokenObj) => {
+      const underlyingToken = tokenObj.token.underlyingTokens[0];
+      finalResponse[`${tokenObj.token.chainId}-${underlyingToken.address}-0`] = {
+        underlying: underlyingToken.underlyingTokens[0].address,
```
FiboApe, line 88: are you sure this underlying address is correct? Remember that we switch underlyings on the fe, and I think the api actually returns it as it is on chain

th0rOdinson: Addressed!

FiboApe (approved):

PR #451, opened 2023-09-25
fix(token-picker): show error when importing a token

th0rOdinson:
## Demo
When the user enters an invalid address:
Screenshot 2023-09-25 at 17 20 13

When the user enters a valid address, but no token was found
Screenshot 2023-09-25 at 17 21 02


FiboApe (approved):
Perfect!

PR #459, opened 2023-10-05
fix(*): revert hooks optimization

FiboApe:


th0rOdinson (approved):
Great job, Fibo! My only concern is the additional boilerplate code shared across our custom hooks. We can consider addressing this in the future.

PR #460, opened 2023-10-05
fix(app/root): correctly load lato font

FiboApe:


0xged (approved):
why do u hate me when i give you nothing but love

PR #470, opened 2023-10-17
feat(labels): Implement GET & POST for labels with privy authorization

th0rOdinson:


FiboApe (commented):
Can we also modify the `Address` component to search for labels based on the labelService or accountService?

apps/root/src/services/accountService.ts
```diff
@@ -11,13 +12,13 @@ enum WalletType {
 type Wallet = {
   type: WalletType;
   address: string;
-  label: string;
+  label?: string;
```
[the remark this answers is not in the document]
th0rOdinson: I believe it would be useful to handle `undefined` values. In the UI, we would easily detect if there is no label set and choose to show or hide the label component. Otherwise, we would be showing the same value duplicated (Label + Address) when no label is set

apps/root/src/common/components/address/index.tsx
```diff
@@ -25,6 +27,7 @@ const Address = ({ address, trimAddress, trimSize }: AddressProps) => {
   return (
     <>
       {addressEns ||
+        storedLabels[address] ||
```
FiboApe, line 30: We should prioritize the user stored label to the ens name

apps/root/src/hooks/useStoredLabels.ts
```diff
@@ -0,0 +1,23 @@
...
+
+  const prevLabelsRef = React.useRef<AccountLabels>(storedLabels);
+
+  React.useEffect(() => {
+    const currentLabels = labelService.getStoredLabels();
+    if (!isEqual(prevLabelsRef.current, currentLabels)) {
+      setStoredLabels(currentLabels);
+      prevLabelsRef.current = currentLabels;
+    }
+  }, [labelService]);
```
FiboApe, line 18: we should also add the accountService.getWallets value here, so it refreshes when the wallets/user changes

th0rOdinson: Nice catch! Thank you

apps/root/src/common/components/address/index.tsx
th0rOdinson: Sure! Done 😄

FiboApe (approved):

PR #472, opened 2023-10-19
feat(walletService): Modify walletService to work through accountService

th0rOdinson:


apps/root/src/services/accountService.ts
```diff
@@ -22,12 +22,15 @@ export default class AccountService implements IAccountService {
...
+  setActiveWallet(wallet: string): void {
+    const newActiveWallet = find(this.user?.wallets || [], { address: wallet.toLowerCase() })!;
+    this.activeWallet = newActiveWallet;
+  }
 
-    await this.web3Service.connect(provider, undefined, undefined, this.activeWallet.type === WalletType.embedded);
-    return;
+  async setAndConnectActiveWallet(wallet: string): Promise<void> {
+    this.setActiveWallet(wallet);
+    const provider = await this.activeWallet!.getProvider();
```
FiboApe, line 32: is the ! actually needed?

apps/root/src/services/accountService.ts
```diff
@@ -22,12 +22,15 @@ export default class AccountService implements IAccountService {
...
-    this.activeWallet = find(this.user?.wallets || [], { address: wallet.toLowerCase() })!;
-    const provider = await this.activeWallet.getProvider();
+  setActiveWallet(wallet: string): void {
+    const newActiveWallet = find(this.user?.wallets || [], { address: wallet.toLowerCase() })!;
+    this.activeWallet = newActiveWallet;
+  }
 
-    await this.web3Service.connect(provider, undefined, undefined, this.activeWallet.type === WalletType.embedded);
-    return;
+  async setAndConnectActiveWallet(wallet: string): Promise<void> {
```
FiboApe, line 30: what would be the difference between setting the active wallet and connecting the active wallet? Both should be the same action

PR #474, opened 2023-10-20
feat(AddressComponent): allow modifying a specific address label

th0rOdinson:


apps/root/src/hooks/useLabelHandler.ts
```diff
@@ -0,0 +1,39 @@
+import React from 'react';
```
FiboApe, line 1: Love this label handler!

apps/root/src/services/labelService.ts
```diff
@@ -32,19 +32,59 @@ export default class LabelService {
...
+
+  async deleteLabel(labeledAddress: string): Promise<void> {
+    const user = this.accountService.getUser();
+    if (!user) {
+      return;
+    }
+    const currentLabels = this.labels;
+    if (currentLabels.hasOwnProperty(labeledAddress)) {
+      const deletedLabel = this.labels[labeledAddress];
+      delete currentLabels[labeledAddress];
```
FiboApe, line 75: should this actually be inside the try? since we want to revert all that when it fails right?

FiboApe (approved):
Great work!

apps/root/src/services/labelService.ts
th0rOdinson: You're right, it makes the code clearer to move those lines into the `try` block. Even though they're unlikely to fail due to the `hasOwnProperty` check, I think it's a good practice. I'll move them. Thanks for the feedback!

apps/root/src/hooks/useLabelHandler.ts
th0rOdinson: Thank you sir

PR #475, opened 2023-10-23
feat(contactList): create contactListService and allow to fetch contactList

th0rOdinson:


apps/root/src/services/labelService.ts
```diff
@@ -1,30 +1,36 @@
...
-  async fetchLabels(): Promise<void> {
+  async fetchLabelsAndContactList(): Promise<AccountLabelsAndContactList | undefined> {
     const user = this.accountService.getUser();
     if (!user) {
       return;
     }
+
     try {
-      this.labels = await this.meanApiService.getAccountLabels(user.id);
+      const accountLabelsAndContactList = await this.meanApiService.getAccountLabelsAndContactList(user.id);
```
FiboApe, line 32: you can just return `this.meanApiService.getAccountLabelsAndContactList(user.id)`

FiboApe (approved):
Pretty good!!! Just a minor improvement

apps/root/src/services/labelService.ts
th0rOdinson: Sweet!

PR #483, opened 2023-10-31
Bly 1157

FiboApe:


apps/root/src/frame/components/connect-wallet-option-modal/index.tsx
```diff
@@ -0,0 +1,85 @@
...
+                  <FormattedMessage
+                    description="loginOptionWalletTitle"
+                    defaultMessage="Directly connect your wallet"
+                  />
+                </Typography>
+                <WalletIcon fontSize="inherit" />
+                <Typography variant="body1">
+                  <FormattedMessage
+                    description="loginOptionPrivyDesc"
+                    defaultMessage="You wont get the amazing benefits of multi-wallet but its not privy :shrug:"
```
th0rOdinson, line 73: Shouldn't this new messages be compiled?

apps/root/src/services/meanApiService.ts
```diff
@@ -42,14 +44,22 @@ export default class MeanApiService {
...
+  constructor(
+    contractService: ContractService,
+    axiosClient: AxiosInstance,
+    providerService: ProviderService,
+    accountService: AccountService
+  ) {
     this.axiosClient = axiosClient;
     this.contractService = contractService;
     this.providerService = providerService;
+    this.accountService = accountService;
```
th0rOdinson, line 62: With the implementation of `accountService`, what are your thoughts on removing the `accountId` parameter from the six methods associated with labels and the contact list?

th0rOdinson (approved):

apps/root/src/frame/components/connect-wallet-option-modal/index.tsx
FiboApe: Not really, this text wont be there once we get the ones from the designer

apps/root/src/services/meanApiService.ts
FiboApe: Good thinking! Will implement

PR #489, opened 2023-11-07
feat(contactList): Implement transfer

th0rOdinson:
- feat(contactList): Implement transfer functions for tokens of type native, ERC20 and ERC721
- Implement UI for Transfer
- Refactor token selector to be consumed by transfer and aggregator


apps/root/src/pages/transfer/transfer-container/components/recipient-address/index.tsx
```diff
@@ -0,0 +1,122 @@
...
+  const inputRegex = RegExp(/^[A-Fa-f0-9x]*$/);
+  const { isValidRecipient } = validateAddress(recipientInput, activeWallet?.address);
+
+  React.useEffect(() => {
+    const { isValidRecipient: isValid } = validateAddress(recipientParam || '', activeWallet?.address);
+    if (recipientParam && isValid) {
+      setRecipientInput(recipientParam);
+      dispatch(setRecipient(recipientParam));
+    }
+  }, [dispatch, validateAddress]);
```
FiboApe, line 36: no need to pass validateAddress, but it should change based on activeWallet.address and recipientParam

FiboApe: Or is this for when we load it the first time?

apps/root/src/pages/transfer/transfer-container/components/recipient-address/index.tsx
```diff
@@ -0,0 +1,122 @@
...
+    const { isValidRecipient: isValid } = validateAddress(nextValue, activeWallet?.address);
+
+    if (isValid) {
+      dispatch(setRecipient(nextValue));
+      if (token) {
+        replaceHistory(`/transfer/${currentNetwork.chainId}/${token.address}/${nextValue}`);
+      }
+    } else if (storedRecipient) {
+      // Clear the stored recipient if the input is not valid
+      dispatch(setRecipient(''));
```
FiboApe, line 52: wait if I put an invalid character all the address gets removed?

apps/root/src/pages/transfer/transfer-container/components/recipient-address/index.tsx
```diff
@@ -0,0 +1,122 @@
...
+        margin="normal"
+        spellCheck="false"
+        onChange={(evt) => onRecipientChange(evt.target.value)}
+        inputProps={{
+          pattern: '^0x[A-Fa-f0-9]*$',
+          minLength: 1,
+          maxLength: 79,
+        }}
+        InputProps={{
+          endAdornment: (
```
FiboApe, line 109: Could be cool to also add a paste button here, I think the transfer-to-modal has it

apps/root/src/pages/transfer/transfer-container/components/token-selector/index.tsx
```diff
@@ -0,0 +1,113 @@
...
+  const [fetchedTokenPrice, loadingTokenPrice] = useUsdPrice(
+    selectedToken,
+    parseUnits(amount || '0', selectedToken?.decimals)
+  );
+  const [shouldShowPicker, setShouldShowPicker] = React.useState(false);
+
+  const cantFund = !!selectedToken && !!amount && !!balance && parseUnits(amount, selectedToken.decimals).gt(balance);
+
+  React.useEffect(() => {
+    if (tokenParam) {
```
FiboApe, line 50: why does this need to change when the chainId changes?

apps/root/src/pages/transfer/transfer-container/components/token-selector/index.tsx
```diff
@@ -0,0 +1,113 @@
...
+  const onSetTokenAmount = (newAmount: string) => {
+    if (!selectedToken) return;
+    dispatch(setAmount(newAmount));
+  };
+
+  const addCustomTokenToList = React.useCallback(
+    (customToken: Token) => {
+      dispatch(addCustomToken(customToken));
+    },
+    [selectedNetwork.chainId, dispatch]
```
FiboApe, line 76: no need to change this based on selectedNetwork right?

apps/root/src/pages/transfer/transfer-container/components/transfer-button/index.tsx
```diff
@@ -0,0 +1,46 @@
...
+
+  const parsedAmount = parseUnits(amount || '0', token?.decimals);
+
+  const onTransfer = async () => {
+    if (!recipient || !token || parsedAmount.lte(0) || !activeWallet) {
+      return;
+    }
+    try {
+      const isProtocolToken = token.address === PROTOCOL_TOKEN_ADDRESS;
+      const parsedToken: Token = { ...token, type: isProtocolToken ? TokenType.BASE : TokenType.ERC20_TOKEN };
```
FiboApe, line 24: Mmm I don't think BASE is correct here right? It should be NATIVE

apps/root/src/pages/transfer/transfer-container/components/transfer-button/index.tsx
```diff
@@ -0,0 +1,46 @@
...
+
+      await contactListService.transferTokenToContact({
```
FiboApe, line 26: We should use the transaction modal here for error and loading, no need for success if we are going to show the TransactionConfirmation

apps/root/src/pages/transfer/transfer-container/components/transfer-button/index.tsx
```diff
@@ -0,0 +1,46 @@
...
+        token: parsedToken,
+        amount: parsedAmount,
+      });
+    } catch (e) {
+      console.error(e);
+    }
+  };
+
+  return (
+    <Button variant="outlined" fullWidth onClick={onTransfer} disabled={!recipient || !token || parsedAmount.lte(0)}>
```
FiboApe, line 38: lets move this disabled into its own variable so its clearer

apps/root/src/pages/transfer/transfer-container/components/transfer-form/index.tsx
```diff
@@ -0,0 +1,52 @@
...
+  position: relative;
+  overflow: hidden;
+  border-radius: 20px;
+  flex-grow: 1;
+  background-color: rgba(255, 255, 255, 0.01);
+  backdrop-filter: blur(6px);
+`;
+
+const TransferForm = () => {
+  const {
```
FiboApe, line 25: Maybe we should set all the parameters here, so we don't have to prop drill them to the other components, and they stay a bit more abstracted

apps/root/src/pages/transfer/transfer-container/components/wallet-selector/index.tsx
```diff
@@ -0,0 +1,72 @@
+import React from 'react';
```
FiboApe, line 1: Nice component! It could go directly into `common` since we'll probably use it in other products

apps/root/src/services/conctactListService.ts
```diff
@@ -69,6 +80,55 @@ export default class ContactListService implements IContactListService {
...
+    from,
+    contact,
+    token,
+    amount,
+  }: {
+    from: string;
+    contact: Contact;
+    token: Token;
+    amount: BigNumber;
+  }): Promise<TransactionResponse | undefined> {
```
FiboApe, line 93: this would never return undefined

apps/root/src/pages/transfer/transfer-container/components/transfer-button/index.tsx
FiboApe: We also need to add this to the transactions that we track for the user and build messages for each of the states, check `apps/root/src/hooks/useBuildTransactionDetail.ts` and the other `useBuildTransaction*`

apps/root/src/services/conctactListService.ts
```diff
@@ -69,6 +80,55 @@ export default class ContactListService implements IContactListService {
     }
   }
 
+  async transferTokenToContact({
```
FiboApe, line 83: I'm thinking this method should actually go into the `walletService` since it does not matter if the recipient is a contact or not

FiboApe (commented):
Amazing job here! Left some comments :D

apps/root/src/common/components/token-picker-with-amount/index.tsx
```diff
@@ -48,91 +43,90 @@ const StyledTokenInputContainer = styled.div`
   align-items: stretch;
 `;
 
-type Props = {
-  cantFund: boolean | null;
+type TokenPickerWithAmountProps = {
```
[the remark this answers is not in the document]
th0rOdinson: Sure, I can rename the current 'picker' to be `TokenAmountInput`, and the child component related to the amount can be just `AmountInput`

apps/root/src/pages/transfer/transfer-container/components/recipient-address/index.tsx
th0rOdinson: I tried managing the component state via Redux, but it caused a delay in value updates, impacting the user experience negatively. Now, I only dispatch the value to Redux when it's valid, ensuring the submit method only has a valid or empty value. If Redux has a stored (aways valid) value and the input changes to an invalid value, the stored recipient is reset.

apps/root/src/common/components/token-picker-with-amount/index.tsx
```diff
@@ -48,91 +43,90 @@ const StyledTokenInputContainer = styled.div`
...
-      trackEvent('Aggregator - Set sell order');
-    }
-  };
 
   const onSetMaxBalance = () => {
-    if (balance && from) {
-      if (from.address === PROTOCOL_TOKEN_ADDRESS) {
-        const maxValue = balance.gte(getMinAmountForMaxDeduction(currentNetwork.chainId))
-          ? balance.sub(getMaxDeduction(currentNetwork.chainId))
+    if (balance && selectedToken && currentChainId) {
```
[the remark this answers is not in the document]
th0rOdinson: Niceee!!

apps/root/src/pages/transfer/transfer-container/components/token-selector/index.tsx
th0rOdinson: I used the `chainId` to update the component when the default network was overwritten. Thanks to this comment, I was able to find a better solution by using `tokenParam` as a dependency and only updating Redux when there is no `selectedToken`.

apps/root/src/common/components/token-picker-with-amount/index.tsx
```diff
@@ -48,91 +43,90 @@ const StyledTokenInputContainer = styled.div`
...
           <StyledFormHelperText onClick={onSetMaxBalance}>
             <FormattedMessage
               description="in wallet"
               defaultMessage="Balance: {balance}"
               values={{
-                balance: formatCurrencyAmount(balance, from, 4),
+                balance: formatCurrencyAmount(balance, selectedToken, 4),
               }}
             />
-            <StyledButton onClick={onSetMaxBalance} disabled={isLoadingRoute} color="secondary" variant="text">
```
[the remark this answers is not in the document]
th0rOdinson: My bad, this was not intentional 😅. Thank you!

apps/root/src/pages/transfer/transfer-container/components/recipient-address/index.tsx
th0rOdinson: Yes, this hook is triggered during the initial load. I've made some changes, though, and now everything is consolidated within the `TransferForm`.

apps/root/src/pages/transfer/transfer-container/components/recipient-address/index.tsx
FiboApe: gotcha

apps/root/src/pages/transfer/transfer-container/components/transfer-form/index.tsx
```diff
@@ -0,0 +1,100 @@
...
+    }
+  }, [tokenParam]);
+
+  React.useEffect(() => {
+    if (!recipientParam) return;
+    const { isValidRecipient } = validateAddress(recipientParam, activeWallet?.address);
+    if (isValidRecipient) {
+      dispatch(setRecipient(recipientParam));
+    }
+  }, [dispatch]);
```
FiboApe, line 71: no need to add dispatch

apps/root/src/pages/transfer/transfer-container/components/transfer-form/index.tsx
```diff
@@ -0,0 +1,100 @@
...
+
+  const handleChangeNetworkCallback = React.useCallback(
+    (chainId: number) => {
+      dispatch(setChainId(chainId));
+      replaceHistory(`/transfer/${chainId}`);
+    },
+    [dispatch, replaceHistory]
```
FiboApe, line 78: no need for this dependencies

FiboApe (approved):
Looks good! Very good job hjere!

PR #499, opened 2023-11-29
feat(balances): fetch balances for all connected users

th0rOdinson:
- Improved balance fetching
- Enhanced frequency updates for selected tokens
- Redux structure modification for Web3Service accessibility
- Old FetchBalance calls replaced with Redux store access


apps/root/src/state/balances/balancesUpdater.tsx
```diff
@@ -0,0 +1,19 @@
+// import { IntervalSetActions } from '@constants/timing';
+// import { useAppDispatch } from '@hooks/state';
+// import useInterval from '@hooks/useInterval';
+// import { fetchInitialBalances, fetchPricesForAllChains } from './actions';
+
+const BalancesUpdater = () => {
```
th0rOdinson, line 6: We can discuss about using the api for updating balances, otherwise I can continue this in a new task 😄

apps/root/src/state/balances/hooks.ts
```diff
@@ -0,0 +1,77 @@
...
+
+export function useWalletBalances(
+  walletAddress: string,
+  chainId: number
+): { balances: TokenBalances; isLoadingBalances: boolean; isLoadingPrices: boolean } {
+  const { isLoadingAllBalances, ...allBalances } = useAppSelector((state: RootState) => state.balances);
+  const { balancesAndPrices = {}, isLoadingChainPrices } = allBalances[chainId] || {};
+  const tokenBalances: TokenBalances = {};
+
+  Object.entries(balancesAndPrices).forEach(([tokenAddress, tokenInfo]) => {
```
FiboApe, line 27: this could be a reduce right?

apps/root/src/state/balances/hooks.ts
```diff
@@ -0,0 +1,77 @@
...
+): { balance?: BigNumber; isLoading: boolean } {
+  const allBalances = useAppSelector((state: RootState) => state.balances);
+  if (!token || !walletAddress) {
+    return { balance: undefined, isLoading: false };
+  }
+
+  const chainBalances = allBalances[token.chainId] || {};
+  const isLoading = allBalances.isLoadingAllBalances;
+  const balance =
+    chainBalances?.balancesAndPrices?.[token.address]?.balances?.[walletAddress.toLocaleLowerCase()] ??
```
FiboApe, line 49: chainBalances is always going to exist

apps/root/src/state/balances/hooks.ts
```diff
@@ -0,0 +1,77 @@
...
+
+export function useTokensBalances(
+  tokens: Token[] | null,
+  walletAddress: string,
+  chainId: number
+): { balances: Record<string, BigNumber>; isLoadingBalances: boolean } {
+  const allBalances = useAppSelector((state: RootState) => state.balances);
+
+  tokens?.forEach((token) => {
+    if (token.chainId !== chainId) {
```
FiboApe, line 63: why is this necessary?

apps/root/src/state/balances/reducer.ts
```diff
@@ -0,0 +1,70 @@
...
+          balances: tokenBalance.balances,
+        };
+      });
+    })
+    .addCase(fetchPricesForChain.fulfilled, (state, { payload: { chainId, prices } }) => {
+      Object.entries(prices).forEach(([address, price]) => {
+        state[chainId].balancesAndPrices[address].price = price.price;
+      });
+    })
+    .addCase(setLoadingChainPrices, (state, { payload: { chainId, isLoading } }) => {
```
FiboApe, line 67: here for example you can use `fetchPricesForChain.pending`

FiboApe: Same for the others and setting the loading boolean into true

FiboApe (commented):
God dang amazing work here, lets talk through the comments :)

apps/root/src/pages/dca/create-position/components/step1/components/amount-input/index.tsx
```diff
@@ -20,8 +22,11 @@ type Props = {
 };
 
 const AmountInput = ({ cantFund, balance, fromValueUsdPrice, handleFromValueChange }: Props) => {
+  const activeWallet = useActiveWallet();
   const { fromValue, from } = useCreatePositionState();
 
+  useAutoFetchTokenBalance({ token: from, walletAddress: activeWallet?.address });
```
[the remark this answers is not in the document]
FiboApe: Ok so after having the full context of the PR, I'm not too sure about this approach, I think something cool, would be for the useTokenBalance hook to accept a boolean parameter to see if it should update the token balance or not, what do you think about it?

This hooks with no return value can very rapidly start poluting the code

apps/root/src/state/balances/balancesUpdater.tsx
th0rOdinson: Done

apps/root/src/state/balances/hooks.ts
th0rOdinson: It isn't 😄 , I've now adapted the hook to accept tokens from many chains, and return balances for each of them.

apps/root/src/state/balances/actions.ts
```diff
@@ -0,0 +1,133 @@
...
+  return { chainId, tokenBalances: tokenAccountBalances };
+});
+
+export const fetchPricesForChain = createAsyncThunk<
+  { chainId: number; prices: CurrentPriceForChainResponse },
+  { chainId: number },
+  { extra: ExtraArgument }
+>('prices/fetchPricesForChain', async ({ chainId }, { extra: { web3Service }, getState, dispatch }) => {
+  const sdkService = web3Service.getSdkService();
+  const state = getState() as RootState;
```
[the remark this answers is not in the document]
th0rOdinson: This is a tricky situation because we've defined the `RootState` type based on the return value of `getState`. In this case we're trying to define the `getState` type using `RootState`, leading to a circular dependency issue. Despite spending some time trying to resolve this, I haven't found a solution yet. Perhaps we should consider moving on and address it in the future?

apps/root/src/hooks/useInterval.ts
```diff
@@ -17,7 +17,7 @@ const useInterval = (callback: Function, delay?: number | null) => {
     }
 
     return undefined;
-  }, [delay]);
+  }, [...(customDeps || []), delay]);
```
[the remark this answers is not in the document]
FiboApe: right and what about adding the function as a dependency if we create it with React.useCallback we might bypass it re-rendering everytime ?

apps/root/src/hooks/useAutoFetchTokenBalance.ts
```diff
@@ -0,0 +1,22 @@
...
+  token: Token | null;
+  walletAddress?: string;
+}) {
+  const dispatch = useAppDispatch();
+  const fetchAndUpdateTokens = React.useCallback(async () => {
+    if (token && walletAddress) {
+      await dispatch(updateTokens({ tokens: [token], chainId: token.chainId, walletAddress }));
+    }
+  }, [token, walletAddress]);
+  useInterval(fetchAndUpdateTokens, IntervalSetActions.selectedTokenBalance, [token, walletAddress]);
```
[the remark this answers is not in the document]
FiboApe: Right it should, but I can't remember right now if it did

apps/root/src/state/balances/actions.ts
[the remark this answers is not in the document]
FiboApe: Sure, let me try it tho

apps/root/src/state/balances/actions.ts
```diff
@@ -1,121 +1,130 @@
...
-  try {
-    balances = await sdkService.getMultipleBalances(tokens, walletAddress);
-  } catch (e) {
-    console.error(e);
-  }
+  const balances = await sdkService.getMultipleBalances(tokens, walletAddress);
 
-  const tokenAccountBalances: TokenBalancesAndPrices = {};
-  Object.entries(balances[chainId]).forEach(([tokenAddress, balance]) => {
+  const tokenBalances = Object.entries(balances[chainId]).reduce((acc, [tokenAddress, balance]) => {
```
FiboApe, line 20: you can use `reduce<TokenBalancesAndPrices>` and will have no need for the bottom `as` plus it will typecheck the reducer function return value

apps/root/src/state/balances/actions.ts
```diff
@@ -1,121 +1,130 @@
...
+  (_, { getState, dispatch }) => {
     const state = getState() as RootState;
     const chainsWithBalance = Object.keys(state.balances);
 
-    const pricePromises = chainsWithBalance.map(async (chainId) =>
+    const pricePromises = chainsWithBalance.map((chainId) =>
       dispatch(fetchPricesForChain({ chainId: Number(chainId) }))
     );
-    await Promise.all(pricePromises);
+    void Promise.all(pricePromises);
```
FiboApe, line 68: If we want the pending to be correctly calculated, we should await this promises

apps/root/src/state/balances/actions.ts
```diff
@@ -1,121 +1,130 @@
...
+  const chainIds = Object.values(NETWORKS).map((network) => network.chainId);
+  const wallets = accountService.getWallets().map((wallet) => wallet.address);
+  const state = getState() as RootState;
+  const allTokens = flatten(Object.values(state.tokenLists.byUrl).map((list) => list.tokens));
 
-    const accountBalancesResponse = await meanApiService.getAccountBalances({
-      wallets,
-      chainIds,
-    });
+  const tokenListByChainId = allTokens.reduce(
```
FiboApe, line 84: Carefull I don't think this will actually take into account the `priority` parameter on the tokenLists

apps/root/src/state/balances/hooks.ts
```diff
@@ -22,56 +26,75 @@ export function useWalletBalances(
...
   const { isLoadingAllBalances, ...allBalances } = useAppSelector((state: RootState) => state.balances);
   const { balancesAndPrices = {}, isLoadingChainPrices } = allBalances[chainId] || {};
-  const tokenBalances: TokenBalances = {};
 
-  Object.entries(balancesAndPrices).forEach(([tokenAddress, tokenInfo]) => {
-    const balance = tokenInfo.balances[walletAddress] || BigNumber.from(0);
+  const tokenBalances: TokenBalances = Object.entries(balancesAndPrices).reduce((acc, [tokenAddress, tokenInfo]) => {
+    const balance: BigNumber | undefined = tokenInfo.balances[walletAddress];
     const price = !isNil(tokenInfo.price) ? parseUnits(tokenInfo.price.toFixed(18), 18) : undefined;
     const balanceUsd = price && balance.mul(price);
```
FiboApe, line 33: balance could be undefined here right?

FiboApe (commented):
Just some very minor comments!

apps/root/src/hooks/useInterval.ts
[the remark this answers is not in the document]
th0rOdinson: This is an excellent idea 🏆

apps/root/src/index.tsx
```diff
@@ -44,11 +45,12 @@ const BalancesInitializer = () => {
   const dispatch = useAppDispatch();
   const accountService = useAccountService();
   const wallets = accountService.getWallets();
+  const tokenListByChainId = useTokenListByChainId();
   const [shouldFetch, setShouldFetch] = React.useState(true);
 
   React.useEffect(() => {
     const fetchBalancesAndPrices = async () => {
-      await dispatch(fetchInitialBalances());
+      await dispatch(fetchInitialBalances({ tokenListByChainId }));
```
FiboApe, line 53: are we sure the token lists are all fetched by this point?

FiboApe: And should this also be a dependency for the useEffect?

FiboApe (approved):

PR #500, opened 2023-12-01
feat: enables gnosis chain

0xged:
Enables Gnosis Chain on Mean's frontend.


0xged:
@FiboApe for some reason it doesn't show all the positions. Token names on Gnosis Chain are (really) weird, so I think it might be that. 
You can check on the subgraph: https://thegraph.com/hosted-service/subgraph/mean-finance/dca-v2-yf-gnosis there is 4 positions. Also the FE on the network brings 4 positions. I only see 3 😢 
image

FiboApe (approved):

th0rOdinson (approved):

PR #501, opened 2023-12-01
feat: updates sdk and removes wido

0xged:


0xged:
> I dont see any removal of wido but guessing it is from the update of the sdk version

It does :) - tested locally.

PR #505, opened 2023-12-07
feat(balances): update balances after transaction confirmation

th0rOdinson:


apps/root/src/common/components/transaction-confirmation/index.tsx
```diff
@@ -217,7 +217,7 @@ const TransactionConfirmation = ({ shouldShow, handleClose, transaction, to, fro
         transactionReceipt?.type === TransactionTypes.swap
       ) {
         walletService
-          .getBalance(PROTOCOL_TOKEN_ADDRESS, transactionReceipt?.typeData.transferTo || transactionReceipt.from)
+          .getBalance(transactionReceipt?.typeData.transferTo || transactionReceipt.from, PROTOCOL_TOKEN_ADDRESS)
```
FiboApe, line 220: did this actually change?

apps/root/src/state/transactions/transactionUpdater.tsx
```diff
@@ -234,6 +277,15 @@ export default function Updater(): null {
                 TransitionComponent: Zoom,
               }
             );
+
+            // TODO: What about transfering to own accounts?
```
FiboApe, line 281: how you mean?

apps/root/src/common/components/transaction-confirmation/index.tsx
th0rOdinson: There was an error that happened when updating balances after a swap. It wasn't related to this task, but while debugging my feature, I found out the problem was due to the parameters being in the wrong order.

apps/root/src/state/transactions/transactionUpdater.tsx
th0rOdinson: Oops! I forgot about that.
I thought about a corner case involving transferring or swapping+transferring to one's own wallets. We would need to update the `to` token for the recipientAddress, if that address is loaded as one of the wallets

apps/root/src/common/components/transaction-confirmation/index.tsx
FiboApe: nice

FiboApe (approved):

PR #507, opened 2023-12-11
fix(web3Service): refactor on services dependencies

th0rOdinson:


apps/root/src/services/campaignService.ts
```diff
@@ -21,7 +21,8 @@ export default class CampaginService {
   }
 
   async getCampaigns(userAddress: string): Promise<CampaignsWithoutToken> {
-    const rawCampaigns = await this.meanApiService.getCampaigns(userAddress);
+    const provider = await this.providerService.getSigner(userAddress);
```
FiboApe, line 24: we actually just need a provider, we could use the ones provided by the sdk, in case the specific userAddress doesn't have a provider available (like having your wallet linked but it is not available bc it was added on another device)

th0rOdinson: Absolutely!

FiboApe (approved):

PR #508, opened 2023-12-12
home(*): implement home frame and routing + Portfolio  breakdown

th0rOdinson:
## Demo

https://github.com/Mean-Finance/dca-fe/assets/144699868/afb488d4-02f4-42dc-837c-2b21bc358df9

### DIFF
--- a/apps/root/src/common/components/wallet-selector/index.tsx
+++ b/apps/root/src/common/components/wallet-selector/index.tsx
@@ -56,20 +56,14 @@ type StatePropsDefined = {
   selectedWalletOption: string;
 };
 
-type WalletSelectorProps =
-  | WithAllWalletsOption
-  | WithSetActiveWalletTrue
-  | WithSetActiveWalletFalse
-  | StatePropsDefined;
+type WalletSelectorProps = {
+  options: WithAllWalletsOption | WithSetActiveWalletTrue | WithSetActiveWalletFalse | StatePropsDefined;
+};
 
 export const ALL_WALLETS = 'allWallets';
 
-const WalletSelector = ({
-  allowAllWalletsOption,
-  setSelectionAsActive,
-  onSelectWalletOption,
-  selectedWalletOption,
-}: WalletSelectorProps) => {
+const WalletSelector = ({ options }: WalletSelectorProps) => {
+  const { allowAllWalletsOption, onSelectWalletOption, selectedWalletOption, setSelectionAsActive } = options;
   const user = useUser();
   const activeWallet = useActiveWallet();
   const accountService = useAccountService();
--- a/apps/root/src/common/components/wallet-selector/index.tsx
+++ b/apps/root/src/common/components/wallet-selector/index.tsx
@@ -28,25 +28,48 @@ const StyledMenuItem = styled(MenuItem)`
   gap 5px; 
 `;
 
-interface BaseProps {}
-
-interface WithAllWallets {
-  allowAllWalletsOption: boolean;
+type WithAllWalletsOption = {
+  allowAllWalletsOption: true;
+  setSelectionAsActive?: never;
   onSelectWalletOption: (newWallet: string) => void;
   selectedWalletOption: string;
-}
+};
 
-interface WithoutAllWallets {
+type WithSetActiveWalletTrue = {
   allowAllWalletsOption?: never;
+  setSelectionAsActive: true;
   onSelectWalletOption?: never;
   selectedWalletOption?: never;
-}
+};
 
-type WalletSelectorProps = BaseProps & (WithAllWallets | WithoutAllWallets);
+type WithSetActiveWalletFalse = {
+  allowAllWalletsOption?: never;
+  setSelectionAsActive: false;
+  onSelectWalletOption: (newWallet: string) => void;
+  selectedWalletOption: string;
+};
+
+type StatePropsDefined = {
+  allowAllWalletsOption?: boolean;
+  setSelectionAsActive?: boolean;
+  onSelectWalletOption: (newWallet: string) => void;
+  selectedWalletOption: string;
+};
+
+type WalletSelectorProps =
+  | WithAllWalletsOption
+  | WithSetActiveWalletTrue
+  | WithSetActiveWalletFalse
+  | StatePropsDefined;
 
 export const ALL_WALLETS = 'allWallets';
 
-const WalletSelector = ({ allowAllWalletsOption, onSelectWalletOption, selectedWalletOption }: WalletSelectorProps) => {
+const WalletSelector = ({
+  allowAllWalletsOption,
+  setSelectionAsActive,
+  onSelectWalletOption,
+  selectedWalletOption,
+}: WalletSelectorProps) => {
   const user = useUser();
   const activeWallet = useActiveWallet();
   const accountService = useAccountService();
@@ -59,7 +82,7 @@ const WalletSelector = ({ allowAllWalletsOption, onSelectWalletOption, selectedW
     },
   });
   const onClickItem = (newWallet: string) => {
-    if (!allowAllWalletsOption) {
+    if (setSelectionAsActive) {
       void accountService.setActiveWallet(newWallet);
     }
     if (onSelectWalletOption) {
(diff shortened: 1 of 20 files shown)


apps/root/src/pages/home/components/portfolio/index.tsx
```diff
@@ -0,0 +1,109 @@
...
+            </TableRow>
+          </TableHead>
+          <TableBody>
+            {Object.entries(balances).map(([tokenKey, tokenInfo]) => {
+              const balanceUsd = tokenInfo.price && tokenInfo.balance * tokenInfo.price;
+              return (
+                <TableRow key={tokenKey}>
+                  <TableCell>
+                    <Typography>
+                      {tokenInfo.balance.toFixed(4)} {tokenInfo.token.symbol}
```
FiboApe, line 64: we should be using the formatCurrencyAmount and not toFixed

apps/root/src/pages/home/frame/index.tsx
```diff
@@ -0,0 +1,93 @@
...
+          const tokenKey = `${tokenInfo.token.chainId}-${tokenAddress}`;
+          newAcc[tokenKey] = {
+            price: tokenInfo.price,
+            token: tokenInfo.token,
+            balance: 0,
+            isLoadingPrice: isLoadingChainPrices,
+          };
+
+          Object.entries(tokenInfo.balances).forEach(([walletAddress, balance]) => {
+            const parsedBalance = parseFloat(formatUnits(balance, tokenInfo.token.decimals));
```
FiboApe, line 29: lets use bignumber so we dont lose accuracy, you can use the `add` function from it

apps/root/src/pages/home/frame/index.tsx
```diff
@@ -0,0 +1,93 @@
...
+import { formatUnits } from '@ethersproject/units';
+import CenteredLoadingIndicator from '@common/components/centered-loading-indicator';
+import Activity from '../components/activity';
+import { isUndefined } from 'lodash';
+
+const HomeFrame = () => {
+  const [selectedWalletOption, setSelectedWalletOption] = React.useState(ALL_WALLETS);
+  const { isLoadingAllBalances, ...allBalances } = useAllBalances();
+
+  const portfolioBalances = React.useMemo(
```
FiboApe, line 15: we should also sort this by most amount of balance in usd

apps/root/src/common/components/wallet-selector/index.tsx
```diff
@@ -23,11 +22,34 @@ const StyledNetworkButtonsContainer = styled.div`
...
+
+interface WithAllWallets {
+  allowAllWalletsOption: boolean;
+  onSelectWalletOption: (newWallet: string) => void;
+  selectedWalletOption: string;
+}
+
+interface WithoutAllWallets {
+  allowAllWalletsOption?: never;
+  onSelectWalletOption?: never;
```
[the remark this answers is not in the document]
th0rOdinson: In the classic implementation of the `WalletSelector`, we depended on the value of `activeWallet`, and it was not necessary to use a locally defined `useState`. I have identified four use cases for this component and have created specific prop types for each. Please let me know your thoughts on this approach!

Also, I've realized that type safety is crucial for this component, as some combinations of types can lead to errors. For example, having both `setSelectionAsActive` set to true and `allowAllWalletsOption` set to true would result in problematic behavior.

apps/root/src/common/components/wallet-selector/index.tsx
```diff
@@ -28,25 +28,48 @@ const StyledMenuItem = styled(MenuItem)`
...
+type WithSetActiveWalletTrue = {
   allowAllWalletsOption?: never;
+  setSelectionAsActive: true;
   onSelectWalletOption?: never;
   selectedWalletOption?: never;
-}
+};
+
+type WithSetActiveWalletFalse = {
+  allowAllWalletsOption?: never;
```
FiboApe, line 46: if you don't want this props, you can just not write them, and then do an join on typescript `|`

FiboApe: Or is that a problem with props?

apps/root/src/pages/home/components/portfolio/index.tsx
```diff
@@ -26,14 +39,50 @@ const StyledAssetLogosContainer = styled.div`
...
-const Portfolio = ({ balances }: PortfolioProps) => {
+const PortfolioBodySkeleton = () => {
+  const skeletonRows = Array.from(Array(5).keys());
+  return (
+    <>
+      {skeletonRows.map((i) => (
+        <TableRow key={i}>
+          <TableCell>
+            <Grid container gap={1} direction="column">
+              <Skeleton variant="rounded" width={150} height={24} />
```
FiboApe, line 58: can't you use the `text` variant? So we don't have to defijne width and height

apps/root/src/pages/home/components/portfolio/index.tsx
```diff
@@ -55,50 +104,55 @@ const Portfolio = ({ balances }: PortfolioProps) => {
...
-                        </StyledNetworkLogoContainer>
-                      </StyledAssetLogosContainer>
-                      <Grid>
-                        <Typography>{tokenInfo.token.symbol}</Typography>
-                        <Typography>{tokenInfo.token.name}</Typography>
+            {isLoadingAllBalances ? (
+              <PortfolioBodySkeleton />
+            ) : (
+              orderedBalances.map((tokenInfo) => {
+                return (
```
FiboApe, line 111: no need for return :). You can just `=> (...component)`

apps/root/src/common/utils/currency.ts
```diff
@@ -40,11 +41,21 @@ const toSignificant = (
...
+  threshold = 0.0001,
   format: object = { groupSeparator: '' }
 ): string => {
+  if (isUndefined(currency)) {
+    return '-';
+  }
+
   const quotient = new Decimal(currency).toSignificantDigits(significantDigits);
+
+  if (quotient.lessThan(threshold)) {
```
FiboApe, line 55: this should be already handled by formatCurrencyAmount

th0rOdinson: The `formatCurrencyAmount` function needs a BigNumber. But price and balanceUsd are decimal numbers, so changing them to BigNumber just to fix their decimals looks hard. Maybe we can change formatCurrencyAmount to take strings too? I solve this in this PR by using a different function for strings

apps/root/src/common/components/wallet-selector/index.tsx
th0rOdinson: Every prop should be defined across all union types or it would not be recognized by the component.
https://stackoverflow.com/questions/58630750/react-typescript-component-with-two-different-prop-interfaces

apps/root/src/pages/home/components/portfolio/index.tsx
th0rOdinson: Definitely! Thank you. Font sizes can be adjusted later based on the design

apps/root/src/common/components/wallet-selector/index.tsx
th0rOdinson: Sooo nice!!!! tytyty

apps/root/src/frame/components/navigation/index.tsx
```diff
@@ -72,7 +72,13 @@ const Navigation = ({
   const changeLanguage = useChangeLanguage();
 
   React.useEffect(() => {
-    if (location.pathname === '/positions') {
+    if (location.pathname === '/') {
```
FiboApe, line 75: missing adding the section on the bottom! the icon is the `DashboardIcon`

apps/root/src/constants/routes.tsx
```diff
@@ -1,7 +1,12 @@
 import React from 'react';
 import { defineMessage } from 'react-intl';
-import { DcaInvestIcon, SwapIcon, TransferIcon } from 'ui-library';
+import { DashboardIcon, DcaInvestIcon, SwapIcon, TransferIcon } from 'ui-library';
 
+export const DASHBOARD_ROUTE = {
+  label: defineMessage({ description: 'dashboard', defaultMessage: 'Dashboard' }),
+  key: '',
```
FiboApe, line 7: key should be 'home'

FiboApe (approved):

PR #509, opened 2023-12-13
feat(*): adds navigation

FiboApe:


packages/ui-library/src/components/navigation/help-menu.tsx
th0rOdinson, line 1: Seems to be empty

apps/root/src/frame/components/navigation/index.tsx
```diff
@@ -0,0 +1,176 @@
...
+  const currentRoute = useCurrentRoute();
+  const intl = useIntl();
+  const activeWallet = useActiveWallet();
+  const mode = useThemeMode();
+  const selectedLanguage = useSelectedLanguage();
+  const changeLanguage = useChangeLanguage();
+
+  React.useEffect(() => {
+    if (location.pathname === '/positions') {
+      dispatch(changeRoute('positions'));
```
th0rOdinson, line 76: I think we should decouple routes from tabs reducer, using `react-roter-dom` functions. We would simplify every redirect with just one operation. And hooks like `useCurrentRoute could just be replaced by implementing `useLocation`. What do you think?

th0rOdinson (commented):
This looks sooooo good 🚀

apps/root/src/frame/components/navigation/index.tsx
FiboApe: good idea

th0rOdinson (approved):

PR #510, opened 2023-12-13
feat(transfer): Dont allow sending tokens to token addresses

th0rOdinson:


apps/root/src/hooks/useValidateTransferRecipient.ts
```diff
@@ -0,0 +1,71 @@
...
+          description: 'errorSameAddress',
+        }),
+      });
+      return;
+    }
+
+    if (tokenList[recipient]) {
+      setValidationResult({
+        isValidRecipient: false,
+        errorMessage: intl.formatMessage({
```
FiboApe, line 50: we probably want to use `defineMessage` before `formatMessage`

FiboApe (approved):

PR #514, opened 2023-12-15
fix(balances): fetch token data if not found in tokenList

th0rOdinson:


apps/root/src/pages/home/components/portfolio/index.tsx
```diff
@@ -160,10 +160,9 @@ const Portfolio = ({ balances, isLoadingAllBalances }: PortfolioProps) => {
                       <StyledNetworkLogoContainer>
                         <TokenIcon
                           size="14px"
-                          token={{
-                            ...tokenInfo.token,
```
FiboApe, line 164: would this token not exist? we do use the address and the chain id to search for a logo in case the ghTokenListLogoUrl does not exist

apps/root/src/state/balances/actions.ts
```diff
@@ -65,10 +69,40 @@ export const fetchPricesForAllChains = createAppAsyncThunk<void, void>(
...
+    const tokenContract = await web3Service.contractService.getERC20TokenInstance(chainId, tokenAddress, walletAddress);
+    const [name, symbol, decimals] = await Promise.all([
+      tokenContract.name(),
+      tokenContract.symbol(),
+      tokenContract.decimals(),
+    ]);
+
+    const customToken = toToken({
+      address: tokenAddress,
+      name: capitalizeFirstLetter(name),
```
FiboApe, line 90: no need to capitalize first letter, we should use the one provided on the contract

apps/root/src/state/balances/actions.ts
```diff
@@ -65,10 +69,40 @@ export const fetchPricesForAllChains = createAppAsyncThunk<void, void>(
   }
 );
 
+const fetchTokenDetails = createAppAsyncThunk<
+  Token,
+  { tokenAddress: string; chainId: number; walletAddress: string; tokenList: TokenList }
+>(
+  'balances/fetchTokenDetails',
+  async ({ tokenAddress, chainId, walletAddress, tokenList }, { dispatch, extra: { web3Service } }) => {
+    if (tokenList[tokenAddress]) {
```
FiboApe, line 78: is this already chainSpecific? bc we could be fetching a token from another chain if it has the same address

apps/root/src/pages/home/components/portfolio/index.tsx
th0rOdinson: In order to render a Network logo, I found out that passing the tokenAddress could render an unexpected icon. 
Just passing an empty token with logoUri and it's chainId was a way to make sure about this result

apps/root/src/state/balances/actions.ts
th0rOdinson: This `tokenList` is from a specific chain. That is handled by `fetchInitialBalances` when it calls this function

apps/root/src/pages/home/components/portfolio/index.tsx
FiboApe: Ah gotcha that is the network icon, mmm I'm thinking we should compononentize this

apps/root/src/state/balances/actions.ts
FiboApe: Perfect

apps/root/src/pages/home/components/portfolio/index.tsx
FiboApe: I'll create a ticket for it

FiboApe: Created BLY-1425 and moved it to the next cycle

FiboApe (approved):

PR #516, opened 2023-12-19
feat(wallet-selector): Implemented new UI

th0rOdinson:


packages/ui-library/src/components/svgicon/index.tsx
```diff
@@ -1,4 +1,16 @@
...
 import type { SvgIconProps } from '@mui/material/SvgIcon';
 import SvgIcon from '@mui/material/SvgIcon';
 
-export { SvgIcon, type SvgIconProps };
+interface CustomIconProps extends SvgIconProps {
+  viewBox: string;
+  size?: string;
+}
+
+const CustomSvgIcon = ({ viewBox, size, children, ...props }: CustomIconProps) => (
```
FiboApe, line 10: why did we need this change?

packages/ui-library/src/icons/emptyWallet.tsx
```diff
@@ -0,0 +1,41 @@
...
+        fill="#06AC54"
```
FiboApe, line 13: remember to remove the `fills` on all the new icons! if not they are always going to have the same color :)

packages/ui-library/src/components/icon-menu/index.tsx
```diff
@@ -18,17 +46,27 @@ type IconMenuProps = {
...
 };
 
-const IconMenu = ({ options, color = 'info', variant = 'text', icon, size = 'small' }: IconMenuProps) => {
+const IconMenu = ({
+  options,
+  color = 'info',
+  variant = 'text',
+  icon,
+  size = 'small',
+  blockMenuOpen,
```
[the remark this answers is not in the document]
th0rOdinson: In this specific scenario, when a user edits the address label, it is important that the menu remains closed. This flag is implemented to ensure that the menu will not reopen until the user has finished editing and the label's text field is no longer in focus.

packages/ui-library/src/components/icon-menu/index.tsx
```diff
@@ -63,13 +101,32 @@ const IconMenu = ({ options, color = 'info', variant = 'text', icon, size = 'sma
...
+            onClick,
+            icon: itemIcon,
+            label,
+            secondaryLabel,
+            control,
+            closeOnClick = true,
+            bottomDivider,
+            color: itemColor,
+          }) => (
+            <Box key={label}>
```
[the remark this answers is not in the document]
th0rOdinson: I needed to render `<Divider />` alongside `<MenuItem />` as siblings. At first, I tried to use a React Fragment, but this was unsupported by Material-UI's Menu. Consequently, wrapping them in a plain div became the most viable solution, especially since each pair needed to be returned as a single component inside a map function.

packages/ui-library/src/components/svgicon/index.tsx
th0rOdinson: For cases like the `ContentCopyIcon`, we are migrating to a custom svg icon. The native Mui <SvgIcon /> wrapper, wasn't supporting custom props such as `onClick`. This structure allows to send any kind of prop and it should be properly included in the icon.

packages/ui-library/src/icons/emptyWallet.tsx
th0rOdinson: Good catch! Thank you

apps/root/src/common/components/wallet-selector/index.tsx
```diff
@@ -84,55 +92,103 @@ const WalletSelector = ({ options }: WalletSelectorProps) => {
...
-              </>
-            )}
-          </ConnectButton.Custom>
-        </Select>
-      </StyledNetworkButtonsContainer>
-    </StyledNetworkContainer>
+  return (
+    <IconMenu
+      options={menuOptions}
+      icon={<Typography>{selectedOptionLabel}</Typography>}
```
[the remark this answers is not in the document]
th0rOdinson: I propose generalizing this component to also accept a string as mainDisplay

apps/root/src/common/components/token-picker-modal/components/token-picker/index.tsx
```diff
@@ -336,7 +336,7 @@ const RawRow = ({
           </Typography>
           <Typography variant="body" component="span" sx={{ display: 'flex' }}>
             <Tooltip title={token.address} arrow placement="top">
-              <StyledCopyIcon fontSize="inherit" onClick={onCopyAddress} />
+              <StyledCopyIcon onClick={onCopyAddress} />
```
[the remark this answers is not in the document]
th0rOdinson: It needs it 🌝

packages/ui-library/src/components/icon-menu/index.tsx
```diff
@@ -3,13 +3,41 @@ import { ButtonProps, Button } from '../button';
...
+  ${({ theme: { spacing } }) => `
+  display: flex;
+  align-items: center;
+  gap: ${spacing(2)};
+`}
+`;
+
+const StyledDivider = styled(Divider)`
+  ${({ theme: { spacing } }) => `
+  padding-top: ${spacing(2)};
```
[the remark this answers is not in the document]
th0rOdinson: For this solution, the divider is contained in the same box as the MuiMenuItem, so it needs a made-up gap of 8 px represented by this padding-top. I believe <Divider /> is used in many places and it may not need this gap

packages/ui-library/src/components/icon-menu/index.tsx
```diff
@@ -63,13 +101,32 @@ const IconMenu = ({ options, color = 'info', variant = 'text', icon, size = 'sma
...
+            label,
+            secondaryLabel,
+            control,
+            closeOnClick = true,
+            bottomDivider,
+            color: itemColor,
+          }) => (
+            <Box key={label}>
+              <MenuItem onClick={() => handleCloseWithAction(closeOnClick, onClick)} style={{ color: itemColor }}>
+                <StyledMenuItemContent>
```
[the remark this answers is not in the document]
th0rOdinson: I did a workaround to archieve this by targeting the first Box child in the muiItem

packages/ui-library/src/components/icon-menu/index.tsx
```diff
@@ -63,13 +101,32 @@ const IconMenu = ({ options, color = 'info', variant = 'text', icon, size = 'sma
...
-        ))}
+        {options.map(
+          ({
+            onClick,
+            icon: itemIcon,
+            label,
+            secondaryLabel,
+            control,
+            closeOnClick = true,
+            bottomDivider,
```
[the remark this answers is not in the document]
th0rOdinson: This comment has been enlightening and led me to rethink the current structure :) In particular, I've reconsidered how I was handling the divider and the Box Wrapper for the MenuItem. Thanks!

packages/ui-library/src/components/icon-menu/index.tsx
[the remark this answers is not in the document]
th0rOdinson: I have removed the Box Container now, by a condition based on the OptionType

packages/ui-library/src/components/icon-menu/index.tsx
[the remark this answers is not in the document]
FiboApe: but why woudl the menu reopen while the user is modifying the label?

packages/ui-library/src/components/icon-menu/index.tsx
```diff
@@ -48,7 +84,7 @@ const IconMenu = ({ options, color = 'info', variant = 'text', icon, size = 'sma
   return (
     <div>
       <Button variant={variant} color={color} size={size} onClick={handleClick} endIcon={<KeyboardArrowDownIcon />}>
-        {icon}
+        {typeof mainDisplay === 'string' ? <Typography>{mainDisplay}</Typography> : mainDisplay}
```
FiboApe, line 87: we should set a variant for this typography

packages/ui-library/src/components/icon-menu/index.tsx
```diff
@@ -63,16 +99,40 @@ const IconMenu = ({ options, color = 'info', variant = 'text', icon, size = 'sma
...
+            return (
+              <MenuItem
+                onClick={() => handleCloseWithAction(closeOnClick, onClick)}
+                style={{ color: itemColor }}
+                key={index}
+              >
+                <Box>
+                  {itemIcon}
+                  <Box sx={{ display: 'flex', flexDirection: 'column' }}>
+                    <Box>{label}</Box>
```
FiboApe, line 122: shouldn't this be a typography component?

packages/ui-library/src/components/icon-menu/index.tsx
```diff
@@ -63,16 +99,40 @@ const IconMenu = ({ options, color = 'info', variant = 'text', icon, size = 'sma
...
+              color: itemColor,
+              control,
+              onClick,
+              icon: itemIcon,
+              secondaryLabel,
+            } = option;
+            return (
+              <MenuItem
+                onClick={() => handleCloseWithAction(closeOnClick, onClick)}
+                style={{ color: itemColor }}
```
FiboApe, line 116: style should not be used on MUI components, they accept an `sx` property. In any case this should be a ButtonProps['color'] right?

packages/ui-library/src/components/icon-menu/index.tsx
```diff
@@ -63,16 +99,40 @@ const IconMenu = ({ options, color = 'info', variant = 'text', icon, size = 'sma
...
+                key={index}
+              >
+                <Box>
```
FiboApe, line 119: I'm not sure I understand the need for boxes inside this component, they don't provide any styling

packages/ui-library/src/components/icon-menu/index.tsx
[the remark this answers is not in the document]
th0rOdinson: That's the thing, If the label is editable, and the user clicks on it to write, the menu would open and the label would lose focus, disabling the edit mode

apps/root/src/common/components/wallet-selector/index.tsx
```diff
@@ -84,56 +86,123 @@ const WalletSelector = ({ options }: WalletSelectorProps) => {
...
+    label: intl.formatMessage(
+      defineMessage({
+        defaultMessage: 'Add Wallet',
+        description: 'addWallet',
+      })
+    ),
+    icon: <EmptyWalletIcon />,
+    onClick: onConnectWallet,
+    control: <AddIcon color="success" />,
+    color: colors[mode].semantic.success,
```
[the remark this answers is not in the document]
th0rOdinson: For components like `<Typography />`, success is not a color option. Passing the hex color as a string was the most universal solution I found

packages/ui-library/src/components/icon-menu/index.tsx
th0rOdinson: I have added styles for this box in the components.ts file for all Mui MenuItems. In this case, this box needs to align the Icon with the 1 or 2 labels in the  MenuItem. As a sibling to this box, there is the `control` that need to be positioned at the end to the right

packages/ui-library/src/components/icon-menu/index.tsx
```diff
@@ -3,32 +3,68 @@ import { ButtonProps, Button } from '../button';
...
   closeOnClick?: boolean;
+  color?: string;
 };
 
+type IconMenuOption = DividerOption | MenuOption;
+
 type IconMenuProps = {
   options: IconMenuOption[];
-  icon: React.ReactElement;
+  mainDisplay: React.ReactElement | string;
```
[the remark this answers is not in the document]
th0rOdinson: This approach offers a more general way to handle the component. The `mainDisplay` can be an icon, but it's also capable of being a string for scenarios like the wallet selector

packages/ui-library/src/components/icon-menu/index.tsx
th0rOdinson: I'm changing the prop applied! Thanks 
About ButtonProps, not all of them are applicable to Typography (color=success is not for instance)

packages/ui-library/src/components/icon-menu/index.tsx
FiboApe: Just tried it and we can safely remove this first box

FiboApe: Just add the align items to the MuiMenuItem

packages/ui-library/src/components/icon-menu/index.tsx
FiboApe: We could still send it as the buttonProp and inside the component use a map for the button

th0rOdinson: I found a way to apply most of the button color props to the Typography component
https://github.com/mui/material-ui/issues/29564

I think it's good to go for our current use cases

packages/ui-library/src/components/icon-menu/index.tsx
th0rOdinson: Thank you for giving this a chance! I have moved some styles to the MuiMenuItem global styles as you suggested. Regarding of aligning the items, the `control` icon needs to be positioned at the end of the menuItem. I found a new way to archive this without the <Box> wrapper as it was.

FiboApe (approved):

PR #517, opened 2023-12-20
feat(history): Fetching transactions history

th0rOdinson:


apps/root/src/hooks/useTransactionsHistory.ts
```diff
@@ -0,0 +1,30 @@
...
+  const accountService = useAccountService();
+  const transactionService = useTransactionService();
+  const user = accountService.getUser();
+  const wallets = accountService.getWallets();
+  const storedTransactionsHistory = transactionService.getStoredTransactionsHistory();
+  const fetchingRef = React.useRef(false);
+  React.useEffect(() => {
+    const fetchHistory = async () => {
+      if (fetchingRef.current) return;
+      await transactionService.fetchTransactionsHistory(beforeTimestamp);
```
FiboApe, line 15: this is something that we should do once the user logs in, but not await that promise to finish the login process

FiboApe: since its only going to be done once per session

th0rOdinson: What I tried to do here, is to fetch for the first time, only when the login process has finished (and only fetch if the user is on `/dashboard` or in `/history`).

About the once per session; I thought of two scenarios where we need to re-fetch: 
1. When a user links a new wallet to the account
2. When a wallet last indexed block is far away from the current block

What do you think?

FiboApe: for 1. We wont be refetching probably in the current session since its going to take a while to index that, tho you can check with sam

As for 2. kinda same answer as 1.

If we can load it asynchronously while the user is in another tab we can save more time of the user viewing loading states :)

th0rOdinson: Alright! I can add it inside of the .then() method in web3Service login, in the same way we are loading for labels and contacts. There are some corner cases though, like a user linking a wallet already indexed to a second profile, but we can address those as improvements in the future

FiboApe (approved):
Nice!!!

PR #518, opened 2023-12-20
feat: adds agave as a source of yield

0xged:


apps/root/src/assets/svg/color/index.ts
```diff
@@ -53,6 +54,7 @@ const manifest = {
   YEARN,
   BEEFY,
   VENUS,
+  AGAVE,
```
FiboApe, line 57: this should go to our token list

apps/root/src/assets/svg/color/agave.svg
FiboApe, line 1: Uploaded this to cdn for you QmZfGfaWMdZekDbuTXePqiCM3sx2UeeoaxrfxDmyergCMH

Lets use it through our token list

FiboApe (changes requested):

FiboApe (approved):

PR #523, opened 2023-12-28
feat(ui-library): add transaction receipt

FiboApe:
Screenshot 2023-12-28 at 15 00 52
Screenshot 2023-12-28 at 15 01 01


packages/common-types/src/accountHistory.ts
```diff
@@ -30,4 +31,29 @@ interface ERC20TransferEvent extends BaseEvent {
...
+}
+
+interface BaseEvent extends Omit<BaseApiEvent, 'spentInGas'> {
+  spentInGas: AmountsOfToken;
+  network: NetworkStructWithIcon;
+  explorerLink: string;
+}
+
+export interface ERC20ApprovalEvent extends BaseEvent, Omit<ERC20ApprovalApiEvent, 'token' | 'amount' | 'spentInGas'> {
+  token: TokenWithIcon;
```
th0rOdinson, line 50: What about tokens not registered in our tokenLists? Do you have a default icon in mind, or should we have this icon field as optional?

packages/ui-library/src/components/transaction-receipt/transaction-types-map.ts
```diff
@@ -0,0 +1,13 @@
+import { TransactionEventTypes } from 'common-types';
+import { defineMessage } from 'react-intl';
+
+export const TRANSACTION_TYPE_TITLE_MAP = {
```
th0rOdinson, line 4: We could already implement native transfers, the indexer is already returning those.
- For naming the event:
https://github.com/Mean-Finance/mean-api/blob/dev/src/accounts/handlers/accounts-handler.ts#L170
- For the event interface:
https://github.com/Mean-Finance/mean-api/blob/4ec25a15052a017344def6ba4c55e281db38c824/src/shared/ports/db/repositories/factories/native-repository.ts#L50

packages/ui-library/src/components/transaction-receipt/index.tsx
```diff
@@ -0,0 +1,306 @@
...
+      </StyledSectionContent>
+    </>
+  );
+};
+
+const buildTransactionReceiptForEvent = (transaction: TransactionEvent) => {
+  switch (transaction.type) {
+    case TransactionEventTypes.ERC20_APPROVAL:
+      return <ERC20ApprovalTransactionReceipt transaction={transaction} />;
+      break;
```
th0rOdinson, line 186: We could safely remove break statements due to early returns

th0rOdinson (commented):
This is awesome! Good job

packages/common-types/src/accountHistory.ts
FiboApe: Yup, the TokenIcon already handles that. We should be passing TokenIcon to this. I didn't include token icon into the ui library right now since its tightly packed with the redux fo the token list

packages/ui-library/src/components/transaction-receipt/index.tsx
FiboApe: True

th0rOdinson (approved):
Amazing!

PR #526, opened 2023-12-30
refactor: moves portals to use mean's api

0xged:


apps/root/package.json
```diff
@@ -94,7 +94,7 @@
     "@gnosis.pm/safe-apps-sdk": "^6.1.0",
     "@mean-finance/dca-v2-core": "^3.1.5",
     "@mean-finance/dca-v2-periphery": "^3.3.0",
-    "@mean-finance/sdk": "0.0.167",
+    "@mean-finance/sdk": "0.0.180",
```
0xsambugs, line 97: FYI: I already updated the SDK to `0.0.181` and it's already on `main`

0xged: did you added portals through API too? Just to understand if I merge with main, or close the PR.

0xsambugs: You need to undo changes to `apps/root/package.json` and `yarn.lock`, and only merge the change to `apps/root/src/services/sdkService.ts`

FiboApe (approved):

PR #538, opened 2024-01-16
feat(*): style network selector and create select component

FiboApe:


packages/ui-library/src/components/select/index.tsx
```diff
@@ -1,4 +1,123 @@
...
+        onChange(selectedOption);
+      }
+    },
+    [onChange, options]
+  );
+
+  const handleOnClose = useCallback(() => setSearch(''), []);
+
+  return (
+    <MuiSelect
```
th0rOdinson, line 63: We need to specify icon color (in light mode is not visible): 
Screenshot 2024-01-17 at 16 47 26

th0rOdinson (approved):
Looks pretty nice! Good job

packages/ui-library/src/components/select/index.tsx
FiboApe: Oh good catch

PR #541, opened 2024-01-18
Implement add contact + General styles

th0rOdinson:


apps/root/src/pages/transfer/components/recipient-address/components/add-contact-modal/index.tsx
```diff
@@ -0,0 +1,227 @@
...
+import useContactListService from '@hooks/useContactListService';
+import useStoredContactList from '@hooks/useStoredContactList';
+import useValidateAddress, { ValidationOutput } from '@hooks/useValidateAddress';
+import CenteredLoadingIndicator from '@common/components/centered-loading-indicator';
+import styled from 'styled-components';
+import ErrorCircleIcon from 'ui-library/src/icons/errorCircle';
+
+interface AddContactModalProps {
+  open: boolean;
+  setOpen: React.Dispatch<React.SetStateAction<boolean>>;
```
FiboApe, line 13: We've got the type `SetStateCallback` for this :D

apps/root/src/pages/transfer/components/recipient-address/components/add-contact-modal/index.tsx
```diff
@@ -0,0 +1,227 @@
...
+  goBackToTransfer: () => void;
+}
+
+const StyledStatusTitle = styled(Typography).attrs({ variant: 'h5' })`
+  ${({ theme: { palette } }) => `
+  font-weight: bold;
+  color: ${colors[palette.mode].typography.typo1};
```
FiboApe, line 20: I'm starting to think that we should set this directly on the typography since its being used quite a lot everywhere

apps/root/src/pages/transfer/components/recipient-address/components/add-contact-modal/index.tsx
```diff
@@ -0,0 +1,227 @@
...
+  icon: React.ReactElement;
+  title: React.ReactElement;
+  description: React.ReactElement;
+  button: React.ReactElement;
+}) => (
+  <Grid container direction="column" rowGap={6}>
+    <Grid item xs={12}>
+      {icon}
+    </Grid>
+    <Grid container rowGap={2} direction="column">
```
FiboApe, line 49: every child of a container grid needs to be an item. Take a look at https://mui.com/material-ui/react-grid/ since I'm seeing quite a bit of this

apps/root/src/pages/transfer/components/recipient-address/components/add-contact-modal/index.tsx
```diff
@@ -0,0 +1,227 @@
...
+    };
+  }, [contactList, contactAddress]);
+
+  const { isValidAddress, errorMessage } = useValidateAddress({
+    address: contactAddress,
+    additionalValidations: [validateContactRepetition],
+  });
+
+  const onChangeAddress = (nextValue: string) => {
+    if (!inputRegex.test(nextValue)) {
```
FiboApe, line 99: if we have the useValidateAddress, do we need to parse here?

FiboApe (commented):
I feel like you are using the Grid as the new Box, if you just need a display flex property use a div. A Grid should be used when we want to format elements to be reorganized when screen sizes change

packages/ui-library/src/components/select/index.tsx
```diff
@@ -67,6 +73,7 @@ function Select<T extends { key: string | number }>({
       onChange={handleChangeNetwork}
       onClose={handleOnClose}
       placeholder={placeholder}
+      IconComponent={StyledKeyboardArrowDown}
```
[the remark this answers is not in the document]
th0rOdinson: I did this for two reasons:
1. New design changes the icon for the select to be of this new shape
2. With new styles implemented on this PR, icons inside an input change color based on current state (like startIcon & endIcon). IconComponent for Select should remain always the same color.

apps/root/src/pages/transfer/components/recipient-address/components/add-contact-modal/index.tsx
th0rOdinson: Are you referring to color of h5? Or which style specifically

apps/root/src/pages/transfer/components/recipient-address/components/add-contact-modal/index.tsx
```diff
@@ -0,0 +1,227 @@
...
+    [setOpen]
+  );
+
+  return (
+    <Modal
+      open={open}
+      onClose={() => setOpen(false)}
+      closeOnBackdrop={true}
+      title={
+        !postContactStatus && (
```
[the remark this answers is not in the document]
th0rOdinson: No title at all when status is success or error

apps/root/src/pages/transfer/components/recipient-address/components/add-contact-modal/index.tsx
FiboApe: yeah the colors on the variants of typography, it seems like we are adding the color each time we use them, and they seem to be quite repetitive

apps/root/src/pages/transfer/components/recipient-address/components/contact-list-modal/index.tsx
```diff
@@ -67,65 +80,80 @@ const ContactListModal = ({ shouldShow, setShouldShow, onClickContact }: Contact
...
-      open={shouldShow}
-      onClose={() => setShouldShow(false)}
-      closeOnBackdrop={true}
-      title={<FormattedMessage description="contactList" defaultMessage="Contact list" />}
-      headerButton={
-        contactList.length !== 0 && (
-          <Button variant="outlined" color="primary" onClick={onOpenAddContactForm}>
-            <Typography variant="body1" fontWeight="bold">
+    <>
+      <AddContactModal
```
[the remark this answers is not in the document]
th0rOdinson: To prevent this issue, I added the following condition to the larger 'ContactList' Modal:
```open={open && !openAddContactModal}```
I attempted to merge both modals into a single component, but encountered complications with prop drilling, as I was trying to maintain separation of concerns by keeping everything in distinct components. If you believe this approach isn't completely robust, I'm open to reviewing it further 😄

File: apps/root/src/pages/transfer/components/recipient-address/components/add-contact-modal/index.tsx (hunk 4 above)
Comment (th0rOdinson) [reply apps/root/src/pages/transfer/components/recipient-address/components/add-contact-modal/index.tsx:99]: Mmm I think I can make the input controlled by this hook! I'll try it out

File: packages/ui-library/src/theme/variants/input-base-variants.ts [hunk 8]
```diff
@@ -19,8 +22,32 @@ export const buildInputBaseVariant = (mode: 'light' | 'dark'): Components => ({
       root: {
         borderRadius: SPACING(2),
         backgroundColor: colors[mode].background.secondary,
+        color: colors[mode].typography.typo1,
+        '& .MuiSvgIcon-root': {
+          transition: 'color 200ms',
+          color: colors[mode].typography.typo4,
+        },
+        '&.Mui-focused': {
```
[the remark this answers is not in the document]
th0rOdinson: I tried to do it that way initially, not only styles weren't applied, but also mui threw an error in the console suggesting `Instead, you need to use the '&.Mui-focused' syntax:`

apps/root/src/pages/transfer/components/recipient-address/components/add-contact-modal/index.tsx
th0rOdinson: There are two main cases where colors are already set up by yourself 😄 `body` & `bodySmall`. I've done a little research and didn't find any other common pattern relating Typography variant to color.
I have removed unnecessary color specifications in our codebase, as the variants already cover this for the two cases mentioned

packages/ui-library/src/components/container-box/index.tsx
```diff
@@ -0,0 +1,29 @@
...
+  alignItems?: CSSProperties['alignItems'];
+  alignContent?: CSSProperties['alignContent'];
+  flexWrap?: CSSProperties['flexWrap'];
+  gap?: number;
+  fullWidth?: boolean;
+}
+
+const StyldContainerBox = styled.div<ContainerBoxProps>`
+  display: flex;
+  flex-direction: ${({ flexDirection }) => flexDirection};
```
FiboApe, line 16: we should probably set defaults for all this

packages/ui-library/src/components/container-box/index.tsx
```diff
@@ -0,0 +1,29 @@
...
+  justify-content: ${({ justifyContent }) => justifyContent};
+  align-items: ${({ alignItems }) => alignItems};
+  align-content: ${({ alignContent }) => alignContent};
```
FiboApe, line 19: do we actually use align-content somewhere?

packages/ui-library/src/components/container-box/index.tsx
```diff
@@ -0,0 +1,29 @@
...
+  flex-wrap: ${({ flexWrap }) => flexWrap};
+  width: ${({ fullWidth }) => fullWidth && '100%'};
+  gap: ${({ gap, theme: { spacing } }) => gap && spacing(gap)};
+`;
+
+const ContainerBox = ({ children, ...props }: PropsWithChildren<ContainerBoxProps>) => {
```
FiboApe, line 25: You can just export the styled component

packages/ui-library/src/components/virtualized-table/index.tsx
```diff
@@ -1,26 +1,17 @@
 import React, { forwardRef } from 'react';
 import { Table, TableBody, TableContainer, TableHead, TableRow, Typography, Paper } from '../';
-import { colors } from '../../theme';
 import styled from 'styled-components';
 import { TableVirtuoso, TableComponents, ItemContent, ScrollerProps, FixedHeaderContent } from 'react-virtuoso';
 
 const StyledCellTypography = styled(Typography).attrs({
```
FiboApe, line 6: What about adding the variant name to the cell? `StyledBodyTypography`

FiboApe (approved):

PR #544, opened 2024-01-18
fix(labels): conform to new api structure

FiboApe:


packages/common-types/src/accountLabels.ts
```diff
@@ -1,4 +1,4 @@
-export type AccountLabels = Record<string, string>;
+export type AccountLabels = Record<string, { label: string; lastModified?: number }>;
```
th0rOdinson, line 1: Why is timestamp optional?

FiboApe: A lot of the functions used AccountLabels for posting a label, and only used the `label` part, lastModified was not required for those

th0rOdinson (approved):

packages/common-types/src/accountLabels.ts
0xsambugs: I personally find it confusing to make it optional so that you can use it for both reading and writing. I would personally use two different types in that case, but I'll leave it to you 😄

0xsambugs (approved):

PR #545, opened 2024-01-22
Allow modifying contact from modal

th0rOdinson:


apps/root/src/services/conctactListService.ts
```diff
@@ -1,11 +1,18 @@
...
 import MeanApiService from './meanApiService';
-import { findIndex, remove } from 'lodash';
+import { map } from 'lodash';
 import AccountService from './accountService';
+import WalletService from './walletService';
+import LabelService from './labelService';
+import { EventsManager } from './eventsManager';
 
-export default class ContactListService implements IContactListService {
+export interface ContactListServiceData {
```
FiboApe, line 11: I'm not happy about this, but I cant figure out a better way to do this

apps/root/src/services/conctactListService.ts
```diff
@@ -14,83 +21,104 @@ export default class ContactListService implements IContactListService {
...
   }
 
-  async removeContact(contact: Contact): Promise<void> {
+  async addContact(contact: Contact): Promise<void> {
     const user = this.accountService.getUser();
     if (!user) {
       return;
     }
     const currentContacts = [...this.contactList];
+    const currentLabel = this.labelService.labels?.[contact.address];
```
FiboApe, line 76: why would labels be undefined?

apps/root/src/services/conctactListService.ts
```diff
@@ -14,83 +21,104 @@ export default class ContactListService implements IContactListService {
...
+    this.labelService = labelService;
   }
 
-  async addContact(contact: Contact): Promise<void> {
+  get contactList() {
+    return this.serviceData.contactList;
+  }
+
+  set contactList(contactList) {
+    this.serviceData = { ...this.serviceData, contactList };
```
FiboApe, line 50: Once we get to positionService we are going to need to figure out a better way to handle this, since the setter is not going to active when changing subproperties of the object

apps/root/src/services/conctactListService.ts
```diff
@@ -14,83 +21,104 @@ export default class ContactListService implements IContactListService {
...
     }
   }
 
-  getContacts(): ContactList {
-    return this.contactList;
-  }
+  async initializeAliasesAndContacts(): Promise<void> {
+    const labelsAndContactList = await this.fetchLabelsAndContactList();
+    if (labelsAndContactList) {
+      this.labelService.labels = labelsAndContactList.labels;
```
FiboApe, line 116: shouldn't the label Service handle this?

FiboApe: I'm not keen on services side effects

apps/root/src/services/conctactListService.ts
```diff
@@ -14,83 +21,104 @@ export default class ContactListService implements IContactListService {
...
-        signature,
-      });
+      await this.meanApiService.deleteContact({ contactAddress: contact.address, accountId: user.id, signature });
     } catch (e) {
       this.contactList = currentContacts;
       throw e;
     }
   }
 
-  getContacts(): ContactList {
```
FiboApe, line 89: I think we should still keep this, so we don't `bypass` the underlying data logic through the UI

apps/root/src/services/labelService.ts
```diff
@@ -1,64 +1,45 @@
...
-  ) {
+  constructor(meanApiService: MeanApiService, accountService: AccountService) {
+    super({ labels: {} });
     this.meanApiService = meanApiService;
     this.accountService = accountService;
-    this.contactListService = contactListService;
-    this.walletService = walletService;
   }
 
-  getStoredLabels(): AccountLabels {
```
FiboApe, line 33: same as before

apps/root/src/services/labelService.ts
```diff
@@ -116,22 +98,4 @@ export default class LabelService implements ILabelService {
...
-
-  setWalletsAliases(): void {
-    this.accountService.setWalletsAliases(this.labels, this.ens);
-  }
-
-  async initializeAliasesAndContacts(): Promise<void> {
-    const labelsAndContactList = await this.fetchLabelsAndContactList();
-    if (labelsAndContactList) {
-      this.labels = labelsAndContactList.labels;
-      this.contactListService.setContacts(labelsAndContactList.contacts);
```
FiboApe, line 128: Still not sure why we moved this to contacts

apps/root/src/common/components/address/index.tsx
```diff
@@ -35,19 +35,15 @@ const Address = ({
...
-  onEnableEdit,
+  finishLabelEdition,
   showDetailsOnHover,
 }: AddressProps) => {
   const walletService = useWalletService();
   const storedLabels = useStoredLabels();
   const snackbar = useSnackbar();
   const intl = useIntl();
   const [hovered, setHovered] = React.useState(false);
-  const { newLabelValue, handleBlur, handleFocus, handleChange, isFocus } = useLabelHandler(
```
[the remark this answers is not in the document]
th0rOdinson: I've migrated this logic into the new `EditLabelInput` component. I was repeating code rendering a Textfield and adding those 3 input methods in both `<Address />` and `<ContactItem />`

apps/root/src/pages/transfer/components/recipient-address/components/contact-item/index.tsx
```diff
@@ -85,6 +90,30 @@ const ContactItem = ({ contact, onClickContact, onDeleteContact }: ContactItemPr
     );
   }, []);
 
+  const onClickContactHandler = () => {
+    if (contact.address === activeWallet?.address) {
+      snackbar.enqueueSnackbar(
```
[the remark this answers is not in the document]
th0rOdinson: I thought it would be disadvantageous to allow the user to click on their own wallet as a recipient (having it as a contact). Doing so would close the modal, show the input error state, and then it would need to open the modal again. The validation in the input is always present for cases like pasting the address.
But for now, I've put this aside – we can take a closer look at this corner case later :)

apps/root/src/services/conctactListService.ts
th0rOdinson: Having a single prop in storage, as we do now, this is not entirely necessary, but I liked the idea of having this set up already. Let me know if you have anything in mind!

apps/root/src/services/conctactListService.ts
th0rOdinson: I already faced that issue while testing this 😀. Updating the whole object works here, but it might have significant downsides in the positionService

apps/root/src/services/conctactListService.ts
th0rOdinson: It wouldn't, thanks 😃

apps/root/src/services/labelService.ts
th0rOdinson: I found the following use case:
User has a labeled wallet connected, and then wants to add it as a contact, and when it does, chooses a new label. That label should be updated in labelService, so it's synced with both contactList and accountService's wallets.
Adding, editing or removing a label never affects the contactList. But addContact contactList method has an effect on labels.

apps/root/src/services/conctactListService.ts
th0rOdinson: I'm creating a method so we keep separate services responsibilities

apps/root/src/common/components/address/index.tsx
```diff
@@ -26,7 +26,7 @@ interface AddressProps {
   trimAddress?: boolean;
   trimSize?: number;
   editable?: boolean;
-  finishLabelEdition?: () => void;
+  disableLabelEdition?: () => void;
```
FiboApe, line 29: much better name :D

apps/root/src/pages/transfer/components/recipient-address/components/contact-item/index.tsx
```diff
@@ -178,11 +152,7 @@ const ContactItem = ({ contact, onClickContact, onDeleteContact }: ContactItemPr
                   <>
                     <FormattedMessage description="lastUpdated" defaultMessage="Last Updated" />
                     {': '}
-                    {DateTime.fromMillis(contact.label.lastModified).toLocaleString({
-                      month: 'long',
-                      day: 'numeric',
-                      year: 'numeric',
-                    })}
+                    {DateTime.fromMillis(contact.label.lastModified).toFormat('MMMM dd, yyyy')}
```
FiboApe, line 155: DateTime.DATE_MED is what you want :)

FiboApe: and use toLocaleString so it automatically updates once we change the language

FiboApe: https://moment.github.io/luxon/#/formatting?id=presets

apps/root/src/pages/transfer/components/recipient-address/components/contact-item/index.tsx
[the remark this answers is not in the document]
FiboApe: Cool, we should set this error on the input directly once they finish typing, but we can keep this for hardening

apps/root/src/pages/transfer/components/recipient-address/components/contact-item/index.tsx
th0rOdinson: Awesome! Thanks :D

FiboApe (approved):

PR #549, opened 2024-01-26
fix(aggregator): prevent tx simulation on buy orders when having insufficient balance selling native token

th0rOdinson:


apps/root/src/services/aggregatorService.ts
```diff
@@ -113,25 +113,13 @@ export default class AggregatorService {
...
-      if (shouldValidate) {
-        // If user does not have the balance do not validate tx
-        const balance = await this.walletService.getBalance(from.address);
-
-        if (balance.lt(sellAmount)) {
-          shouldValidate = false;
-          hasEnoughForSwap = false;
-        }
-      }
+    if (takerAddress && shouldValidate && sellAmount && balance.lt(sellAmount)) {
```
FiboApe, line 119: this should check for protocol token or better yet, we have to fetch the balance of the token he wants to use

FiboApe (approved):

PR #550, opened 2024-01-26
Transfer confirmation flow

th0rOdinson:


apps/root/src/hooks/useTransactionsHistory.tsx
```diff
@@ -57,7 +59,11 @@ function useTransactionsHistory(): {
 } {
   const transactionService = useTransactionService();
   const [isHookLoading, setIsHookLoading] = React.useState(false);
-  const { isLoading: isLoadingService, history } = transactionService.getStoredTransactionsHistory();
+
+  const { isLoading: isLoadingService, history } = useServiceEvents<TransactionServiceData, TransactionService>(
+    transactionService,
+    'transactionsHistory'
```
FiboApe, line 65: Talked offline to use the method instead of the property name

apps/root/src/pages/transfer/components/transaction-confirmation/index.tsx
```diff
@@ -34,36 +37,20 @@ const StyledOverlay = styled.div`
...
   `}
 `;
 
-const StyledTitleContainer = styled.div`
-  text-align: center;
-`;
-
-const StyledConfirmationContainer = styled.div`
+const StyledConfirmationContainer = styled(ContainerBox).attrs({ justifyContent: 'center', alignItems: 'center' })`
   align-self: stretch;
```
FiboApe, line 46: we should probably add this properties to the ContainerBox

packages/ui-library/src/components/button/index.tsx
```diff
@@ -1,4 +1,10 @@
+import { CSSProperties } from 'react';
 import type { ButtonProps } from '@mui/material/Button';
-import Button from '@mui/material/Button';
+import MuiButton from '@mui/material/Button';
+import styled from 'styled-components';
+
+const Button = styled(MuiButton)<{ maxWidth?: CSSProperties['maxWidth'] }>`
+  max-width: ${({ maxWidth = '350px' }) => maxWidth};
```
FiboApe, line 7: SPACIIIING

packages/ui-library/src/icons/errorCircle.tsx
```diff
@@ -7,13 +7,13 @@ interface IconProps extends SvgIconProps {
...
 
-export default function ErrorCircleIcon({ size, ...props }: IconProps) {
+export default function ErrorCircleIcon({ size = '162px', ...props }: IconProps) {
   const {
     palette: { mode },
   } = useTheme();
   return (
-    <CustomSvgIcon viewBox="0 0 131 131" style={size ? { fontSize: size } : {}} {...props}>
-      <circle cx="53.2344" cy="53.1855" r="52.5488" fill={colors[mode].semantic.error} />
+    <CustomSvgIcon viewBox="0 0 131 131" style={{ fontSize: size }} {...props}>
```
FiboApe, line 15: to fix!

packages/ui-library/src/icons/successCircle.tsx
```diff
@@ -7,12 +7,12 @@ interface IconProps extends SvgIconProps {
...
 }
 
-export default function SuccessCircleIcon({ size, ...props }: IconProps) {
+export default function SuccessCircleIcon({ size = '162px', ...props }: IconProps) {
   const {
     palette: { mode },
   } = useTheme();
   return (
-    <CustomSvgIcon viewBox="0 0 131 131" style={size ? { fontSize: size } : {}} {...props}>
+    <CustomSvgIcon viewBox="0 0 131 131" style={{ fontSize: size }} {...props}>
```
FiboApe, line 15: to fix!

apps/root/src/pages/transfer/components/confirm-transfer-modal/index.tsx
```diff
@@ -0,0 +1,245 @@
...
+        token.decimals + 18
+      )
+    ).toFixed(2),
+  };
+
+  const nativeCurrencyToken = toToken({
+    ...network.nativeCurrency,
+    logoURI: network.nativeCurrency.logoURI || getGhTokenListLogoUrl(network.chainId, 'logo'),
+  });
+  const mainCurrencyToken = toToken({
```
FiboApe, line 82: wwe do use this some other places I think, maybe we should generalize it?

apps/root/src/hooks/useTransactionsHistory.tsx
th0rOdinson: We are actually returning the useState value and not the class property! I think it's worth keeping it like this, what do you think?

FiboApe: internally it is still using the class property so we should change that

packages/ui-library/src/theme/pallete.ts
```diff
@@ -106,13 +112,19 @@ export const lightModePallete: PaletteOptions = {
     contrastText: colors.light.aqua.aqua900,
   },
   error: {
-    main: colors.light.semantic.error,
+    main: colors.dark.semantic.error.primary,
```
FiboApe, line 115: woops, should be .light right?

packages/ui-library/src/components/container-box/index.tsx
```diff
@@ -18,6 +20,8 @@ const ContainerBox = styled.div<ContainerBoxProps>`
   align-items: ${({ alignItems = 'stretch' }) => alignItems};
   flex-wrap: ${({ flexWrap = 'nowrap' }) => flexWrap};
   flex-grow: ${({ flexGrow = 0 }) => flexGrow};
+  align-self: ${({ alignSelf = 'auto' }) => alignSelf};
+  flex: ${({ flex = '0 1 auto' }) => flex};
```
FiboApe, line 24: `flex: 1` already does that :)

apps/root/package.json
```diff
@@ -133,6 +133,7 @@
     "styled-components": "^5.3.0",
     "toformat": "^2.0.0",
     "tss-react": "^4.9.0",
+    "typed-emitter": "^2.1.0",
```
FiboApe, line 136: are you actually using this?

packages/ui-library/src/theme/pallete.ts
th0rOdinson: Good catch 😅

apps/root/src/hooks/useTransactionsHistory.tsx
th0rOdinson: I've developed a new approach for accessing `servicesData` without directly accessing the property itself. However, I've encountered difficulty in passing a method (such as 'getLabels') as a parameter while maintaining the hook's flexibility for all services. 
This is isolated in the last commit, where I've also removed the useState value since it's no longer utilized if it's not being returned. Let me know if this can be improved! Not sure if this is close to what you had in mind.

apps/root/src/services/transactionService.ts
```diff
@@ -149,27 +153,30 @@ export default class TransactionService extends EventsManager<TransactionService
   }
 
   addTransactionToHistory(tx: TransactionApiEvent) {
+    const transactionsHistory = structuredClone(this.transactionsHistory);
```
FiboApe, line 156: why this and not a spread operator?

apps/root/src/hooks/useServiceEvents.ts
```diff
@@ -16,7 +16,7 @@ function useServiceEvents<ServiceData, ServiceType extends EventsManager<Service
     };
   }, []);
 
-  return serviceData[property];
+  return service.getServiceData()[property];
```
FiboApe, line 19: why is this change necessary?

apps/root/src/services/transactionService.ts
th0rOdinson: I was trying a new approach with deep copying, but it wasn't necessary and gave me more trouble anyway. Now it's a spread operator :)

apps/root/src/hooks/useServiceEvents.ts
```diff
@@ -2,21 +2,27 @@ import React from 'react';
...
+    ? ReturnType<ServiceType[Key]>
+    : ServiceType[Key];
+  const getServicePropertyData = () => (service[property] as () => PropertyValue).bind(service) as PropertyValue;
+
+  const [serviceData, setServiceData] = React.useState<PropertyValue>(getServicePropertyData());
 
   React.useEffect(() => {
     const callbackId = uuidv4();
-    service.setCallback(callbackId, setServiceData);
+    service.setCallback(callbackId, () => setServiceData(getServicePropertyData()));
```
FiboApe, line 19: So in this case getServicePropertyData is actually a function that returns the function from the service, is this actually working?

th0rOdinson: It works! But I understand your point, I'm thinking now that it's more clear if getServicePropertyData, gets the service property data instead of the function that does that 😄

FiboApe (approved):
There we go, now its better :D

PR #552, opened 2024-01-29
chore: update sdk to improve buy orders

0xsambugs:
We are now update the SDK to include [this](https://github.com/Mean-Finance/sdk/commit/7c48e1e664abee0b381f2b6d712a59f45614002f) change. 

Now, when we estimate buy orders with sell orders, we take slippage into account. While this approach might not work 100% of the times, it should work far better than the current approach

### DIFF
--- a/apps/root/src/services/sdkService.ts
+++ b/apps/root/src/services/sdkService.ts
@@ -91,7 +91,6 @@ export default class SdkService {
             type: 'prioritized',
             sources: [
               { type: 'coingecko' },
-              { type: 'portals-fi' },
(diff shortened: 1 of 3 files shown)


apps/root/src/services/sdkService.ts
```diff
@@ -91,7 +91,6 @@ export default class SdkService {
             type: 'prioritized',
             sources: [
               { type: 'coingecko' },
-              { type: 'portals-fi' },
```
0xsambugs, line 94: This source is no longer supported

th0rOdinson (approved):

PR #553, opened 2024-01-29
Apply new structure & styles to Aggregator frame

th0rOdinson:


apps/root/src/pages/aggregator/swap-container/components/advanced-settings/index.tsx
```diff
@@ -0,0 +1,47 @@
...
+import { useThemeMode } from '@state/config/hooks';
+
+const StyledCogContainer = styled.div`
+  position: absolute;
+  top: 0;
+  right: 0;
+  padding: ${({ theme: { spacing } }) => `${spacing(1)} ${spacing(2)}`};
+`;
+
+const StyledIconButton = styled(IconButton)`
```
FiboApe, line 16: cant you make this with `dense` prop?

apps/root/src/pages/aggregator/swap-container/components/step1/index.tsx
```diff
@@ -176,11 +163,9 @@ const SwapFirstStep = React.forwardRef<HTMLDivElement, SwapFirstStepProps>((prop
...
             maxBalanceBtn
           />
-        </StyledContentContainer>
-        <ToggleButton isLoadingRoute={isLoadingRoute} />
-      </Grid>
-      <Grid item xs={12} sx={{ paddingTop: '8px !important' }}>
-        <StyledContentContainer>
+          <ToggleButton isLoadingRoute={isLoadingRoute} />
+        </Grid>
+        <Grid item xs={12} sx={{ paddingTop: '8px !important' }}>
```
FiboApe, line 168: is this padding necessary now?

apps/root/src/pages/aggregator/swap-container/components/swap-settings/index.tsx
```diff
@@ -46,6 +47,8 @@ const StyledOverlay = styled.div`
   padding: 24px;
   display: flex;
   overflow: auto;
+  background-color: ${({ theme: { palette } }) => colors[palette.mode].background.quarteryNoAlpha};
+  border-radius: inherit;
```
FiboApe, line 51: this necessary?

apps/root/src/pages/aggregator/swap-container/components/swap/index.tsx
```diff
@@ -1295,53 +1274,32 @@ const Swap = ({ isLoadingRoute, quotes, fetchOptions, swapOptionsError }: SwapPr
...
+              setShouldShowTransferModal={setShouldShowTransferModal}
               onShowSettings={onShowSettings}
               isApproved={isApproved}
               fromValue={fromValue}
               quotes={quotes}
               fetchOptions={fetchOptions}
               refreshQuotes={refreshQuotes}
               swapOptionsError={swapOptionsError}
+              swapButton={
+                <SwapButton
```
FiboApe, line 1286: this parent component receives most of the same props than the swap button, is there a reason why we are passing it through here and not just render it directly on the first step?

apps/root/src/pages/aggregator/swap-container/components/advanced-settings/index.tsx
th0rOdinson: I'm not seeing dense prop available for IconButton! I've tried with size=small but keeps a little padding

apps/root/src/pages/aggregator/swap-container/components/step1/index.tsx
th0rOdinson: I left this styling as it was, because it keeps the token picker separated as it should be. I agree we should remove it and implement our new theme, but maybe we can address that when we style this component on BLY-1637

apps/root/src/pages/aggregator/swap-container/components/swap-settings/index.tsx
th0rOdinson: No it's not, I had sharp corners overflowing the container but they are gone now 🤔

FiboApe (approved):

PR #554, opened 2024-01-29
chore: updates sdk

0xged:
By updating the SDK, we enable Braindex aggregator for Moonriver, and we also deprecate `portals-fi` which has been deprecated already by the SDK. Tested on my local env, and working sweet.


th0rOdinson (approved):
Looking good!

PR #559, opened 2024-02-05
feat(*): disable GHST yield for dca on polygon

FiboApe:


apps/root/src/constants/common.ts
```diff
@@ -151,6 +151,7 @@ export const DCA_TOKEN_BLACKLIST = [
   '0xc19669a405067927865b40ea045a2baabbbe57f5', // Arbitrum - STAR - Hiding until release
   '0xd3594E879B358F430E20F82bea61e83562d49D48', // Optimism - PSP - Hiding until release
 
+  '0x83c0936d916d036f99234fa35de12988abd66a7f', // Polygon - GHST - Disable bc of aave vault full
```
th0rOdinson, line 154: I find this as the ERC-4626 address (https://polygonscan.com/token/0x83c0936d916d036f99234fa35de12988abd66a7f)

Shouldn't we use the ERC-20 one? This is the one displayed in the token selector for DCA: 
0x385eeac5cb85a38a9a07a70c73e0a3271cfb54a7

th0rOdinson (approved):

PR #560, opened 2024-02-06
Style swap step in aggregator

th0rOdinson:


apps/root/src/common/components/transaction-steps/index.tsx
```diff
@@ -170,6 +171,23 @@ export type TransactionAction =
...
+type TransactionActionProps =
+  | TransactionActionApproveTokenProps
+  | TransactionActionApproveTokenSignProps
+  | TransactionActionWaitForApprovalProps
+  | TransactionActionWaitForSimulationProps
+  | TransactionActionWaitForQuotesSimulationProps
+  | TransactionActionSwapProps
+  | TransactionActionCreatePositionProps;
+
+type CommonTransactionActionProps = Pick<
```
FiboApe, line 183: wouldn't this just be ItemProps?

apps/root/src/common/components/transaction-steps/index.tsx
```diff
@@ -229,6 +250,63 @@ const StyledTransactionStepButtonContainer = styled.div`
...
+  ${({ theme: { palette } }) => `
+  color: ${colors[palette.mode].typography.typo3};
+  `}
+`;
+
+const StyledTickCircleIcon = styled(TickCircleIcon)`
+  color: ${({ theme: { palette } }) => colors[palette.mode].semantic.success.darker};
+`;
+
+const CommonTransactionStepItem = ({
```
FiboApe, line 270: Oh this is cool

apps/root/src/common/components/transaction-steps/index.tsx
```diff
@@ -333,7 +407,7 @@ const buildApproveTokenItem = ({
               >
                 {hasPendingApproval ? waitingForAppvText : isPermit2Enabled ? infiniteBtnText : specificBtnText}
               </Button>
-              {!isPermit2Enabled && (
+              {isPermit2Enabled && (
```
FiboApe, line 410: was this a bug that we had?

apps/root/src/common/components/transaction-steps/index.tsx
```diff
@@ -346,32 +420,41 @@ const buildApproveTokenItem = ({
...
+                  <FormattedMessage description="viewExplorer" defaultMessage="View in explorer" />
+                )}
+              </Button>
+              {!isPendingTransaction && receipt && (
+                <ContainerBox gap={2} alignItems="center">
+                  <StyledTickCircleIcon />
+                  <Typography variant="body" fontWeight="600">
+                    <FormattedMessage
+                      description="approveSellAmount"
+                      defaultMessage="Sell {amount} {symbol}"
```
FiboApe, line 446: Is this correct? Isn't this the approval step?

apps/root/src/common/components/transaction-steps/index.tsx
th0rOdinson: Indeed, after testing this once again, I found out that this button wasn't rendering in the right conditions.
Now, this secondary button (exact amount approval) is only rendered when permit2 is enabled

apps/root/src/common/components/transaction-steps/index.tsx
th0rOdinson: Properties such as `explanation` and `done` props are related to TransactionActionProps and not present in ItemProps. By using TransactionActionProps (extended from ItemProps) I can pick those two and also  `isLast` and `isCurrentStep` from ItemProps.
We could redefine the scope of both interfaces but I found this to be the easiest way to proceed

apps/root/src/common/components/transaction-steps/index.tsx
th0rOdinson: I found this label in the design, and it seemed prudent to me to add it, even though it was on another step. I should have informed you beforehand 😅 
Screenshot 2024-02-08 at 11 20 11

PS: After reviewing it, I think it should be "Approve 20 NEXT"

apps/root/src/common/components/transaction-steps/index.tsx
FiboApe: But ItemProps is only used to extend each actionProp so shouldn't done and explanation be there?

apps/root/src/common/components/transaction-steps/index.tsx
FiboApe: Exact approval should actually appear when permit2 is disabled, now that I'm thinking about it we should always show exact approvals

apps/root/src/common/components/transaction-steps/index.tsx
FiboApe: Yup it should be approve

apps/root/src/common/components/transaction-steps/index.tsx
th0rOdinson: Oh I think I get it now!

apps/root/src/common/components/transaction-steps/index.tsx
th0rOdinson: Agree, maybe I didn't explain myself clearly, I believe this is implemented as expected:

### Permit2 Enabled
Screenshot 2024-02-08 at 16 52 10

### Permit2 Disabled
Screenshot 2024-02-08 at 16 55 18

FiboApe: Gotcha

PR #561, opened 2024-02-07
Add recap data to agg steps

th0rOdinson:


apps/root/src/common/components/token-icon/index.tsx
```diff
@@ -25,8 +25,9 @@ function getLogoURL(logoURI: string) {
   return '';
 }
 
-const TokenIcon = ({ token, isInChip, size }: TokenButtonProps) => {
-  const realSize = size || '28px';
+const TokenIcon = ({ token, isInChip, size = 7 }: TokenButtonProps) => {
+  const { spacing } = useTheme();
+  const realSize = spacing(size);
```
FiboApe, line 30: VERY NICE change 👏 👏 👏 👏 👏

apps/root/src/pages/aggregator/swap-container/components/swap-recap-data/index.tsx
```diff
@@ -0,0 +1,94 @@
...
+            <FormattedMessage description="youPay" defaultMessage="You pay" />
+          </Typography>
+          <ContainerBox gap={2} alignItems="center">
+            <TokenIcon token={selectedRoute.sellToken} size={5} />
+            <Typography variant="body" fontWeight="bold">
+              {formatCurrencyAmount(selectedRoute.sellAmount.amount, selectedRoute.sellToken, 2)}{' '}
+              {selectedRoute.sellToken.symbol}
+            </Typography>
+          </ContainerBox>
+          <Typography variant="bodySmall" textAlign="center" pl={7}>
```
FiboApe, line 39: whats the pl prop?

FiboApe (commented):
This is Fire 🔥 

Only thing is that the swap step should only occupy width  as needed
Screenshot 2024-02-07 at 19 49 16
Screenshot 2024-02-07 at 19 49 10

apps/root/src/pages/aggregator/swap-container/components/swap-recap-data/index.tsx
th0rOdinson: Padding left. 5 from `TokenIcon` + 2 from flex `gap` = 7, so it's aligned with amountInUnits value

FiboApe: Gotcha, instead of doing that since it could create some issues in some browsers due to how px and gaps and px's are calculated. What about spliting the box like this? So the amountInUnits value will ALWAYS be aligned with the usd amount
Untitled drawing (7)

th0rOdinson: Oh my... This new box structure almost made it, but It was hard for me to also align the icon with the amount:
Screenshot 2024-02-08 at 11 35 10

I found a way to achieve this using grid! (I have leveraged Mui Box component, although we were trying to avoid it, but find this so useful: https://mui.com/system/grid/)
Screenshot 2024-02-08 at 15 14 15

Let me know what you think!

FiboApe (approved):
EY1lL_

PR #563, opened 2024-02-08
feat(ui-library): add token input with usd amount

FiboApe:
Just missing figuring out with Designer what to do with the `MATIC` and max/balance items
Screenshot 2024-02-08 at 19 23 36


th0rOdinson (commented):
This is amazing! Can't wait to see it in the site 🤩

packages/ui-library/src/components/token-amount-usd-input/index.tsx
```diff
@@ -0,0 +1,337 @@
...
+
+      if (!internalValue || newInternalValue !== parseFloat(internalValue).toFixed(2)) {
+        setInternalValue(newInternalValue);
+      }
+    } else {
+      throw new Error('invalid inputType');
+    }
+  }, [value]);
+
+  const onChangeType = () => {
```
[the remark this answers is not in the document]
FiboApe: Fixed it!

apps/root/src/common/components/token-selector/index.tsx
```diff
@@ -0,0 +1,107 @@
...
+  const activeWallet = useActiveWallet();
+  const intl = useIntl();
+  const selectedNetwork = useSelectedNetwork();
+  const { balances } = useWalletBalances(activeWallet?.address, selectedNetwork.chainId);
+
+  const tokens = useTokenList({ allowAllTokens: true, filterChainId: true, filter: false });
+
+  const availableTokens = React.useMemo(
+    () =>
+      Object.keys(balances).map<OptionWithKeyAndToken>((tokenAddress) => ({
```
th0rOdinson, line 74: I believe this is working as expected, but when having no balances at all in the selectedNetwork, the the token selector is empty, and I think it can be confusing

Screenshot 2024-02-14 at 10 07 51

The input doesn't have a label or placeholder, so in this case, it's not clear what is this input for. Also, when opening the list, it's just empty.
I'm thinking we could add a default option saying 'You have no balance in the selected network" + Disable input

apps/root/src/common/components/token-selector/index.tsx
```diff
@@ -0,0 +1,107 @@
...
+        ...balances[tokenAddress],
+        key: tokenAddress,
+        token: tokens[tokenAddress],
+      })),
+    [balances]
+  );
+
+  const selectedItem = React.useMemo(
```
th0rOdinson, line 82: We should add a default option

packages/ui-library/src/components/token-amount-usd-input/index.tsx
```diff
@@ -0,0 +1,346 @@
...
+              onChange,
+              nextValue: evt.target.value,
+              decimals: 2,
+            })
+          }
+          value={value || ''}
+          onFocus={onFocus}
+          onBlur={onBlur}
+          startAdornment="$"
+          placeholder="0"
```
th0rOdinson, line 190: Screenshot 2024-02-14 at 09 10 13

I found that autoComplete is making the input look nasty

th0rOdinson (approved):
Amaaazinngggggg 🚀

PR #564, opened 2024-02-09
Merge sign and simulate steps

th0rOdinson:


packages/common-types/src/index.ts
```diff
@@ -91,6 +89,25 @@ export interface TransactionActionApproveTokenData {
...
 
+export enum SignStatus {
+  none = 'none',
+  signed = 'signed',
+  failed = 'failed',
+}
+
+export interface TransactionActionApproveTokenSignDCAData {
+  signStatus: SignStatus;
+  simulation?: BlowfishResponse;
```
FiboApe, line 100: we don't actually simulate on dca :)

packages/common-types/src/index.ts
```diff
@@ -91,6 +89,25 @@ export interface TransactionActionApproveTokenData {
...
+  from: Token;
+  sellAmount: bigint;
```
FiboApe, line 102: this is actually only relevant for swap

apps/root/src/pages/dca/create-position/components/swap/index.tsx
```diff
@@ -637,12 +638,12 @@ const Swap = ({
...
     const newSteps = [...transactions];
 
-    const signIndex = findIndex(transactions, { type: TRANSACTION_ACTION_APPROVE_TOKEN_SIGN });
+    const signIndex = findIndex(transactions, { type: TRANSACTION_ACTION_APPROVE_TOKEN_SIGN_DCA });
 
     if (signIndex !== -1) {
       newSteps[signIndex] = {
         ...newSteps[signIndex],
-        done: true,
+        // done: true,
```
FiboApe, line 646: should we remove this if we are not going to use it?

apps/root/src/pages/aggregator/swap-container/components/swap/index.tsx
```diff
@@ -777,105 +769,90 @@ const Swap = ({ isLoadingRoute, quotes, fetchOptions, swapOptionsError }: SwapPr
...
+                      emptyTokenWithDecimals(18),
+                      3,
+                      2
+                    )
+                  ).toFixed(3);
+
+                if (isThereABetterQuote && (Number(isBetteryBy) > 0 || !originalQuote)) {
+                  dispatch(setSelectedRoute(originalQuote || { ...selectedRoute, willFail: true }));
+                  setBetterQuote(sortedQuotes[0]);
+                  setShouldShowBetterQuoteModal(true);
```
FiboApe, line 811: We should actually remove this, since we are not going to show the modal anymore, we will just update the transaction data that is on top of the steps. We will automatically select a better quote for the user

packages/common-types/src/index.ts
th0rOdinson: I'm seeing amount as a required field for `getPermit2DcaSignedData`. Im renaming this variables to work for DCA and Swap concepts.
Also I have included a SuccessLabel for DCA Sign Step saying 'Invest X TKN' as we do for approve and for sign in Swap

Screenshot 2024-02-14 at 13 30 15

FiboApe: Mmm i think i would put "Token authorization signed" instead of the amount

apps/root/src/pages/aggregator/swap-container/components/swap/index.tsx
th0rOdinson: I tried just removing the modal, but it wouldn't apply the quote selection. We have a task created for this, so I think it's better to handle all this changes in BLY-1642

FiboApe: cool

FiboApe (approved):

PR #570, opened 2024-02-15
feat(swap): Notify better quote inside transaction steps

th0rOdinson:
### Better Quote Notification

https://github.com/Mean-Finance/dca-fe/assets/144699868/69fe35a5-8e5f-4f05-87f3-1b6ee5d1341d

### Notistack Notifications
Screenshot 2024-02-15 at 14 01 58
Screenshot 2024-02-15 at 14 12 31
Screenshot 2024-02-15 at 14 13 55


apps/root/src/pages/aggregator/swap-container/components/swap/index.tsx
```diff
@@ -1180,16 +1181,6 @@ const Swap = ({ isLoadingRoute, quotes, fetchOptions, swapOptionsError }: SwapPr
...
-        selectedRoute={selectedRoute}
-        betterQuote={betterQuote}
-        onCancel={() => setShouldShowBetterQuoteModal(false)}
-        onGoBack={handleBackTransactionSteps}
-        open={shouldShowBetterQuoteModal}
-        onSelectBetterQuote={(response: BlowfishResponse) =>
-          handleTransactionSimulationWait(transactionsToExecute, response)
-        }
-      />
       <FailedQuotesModal
```
FiboApe, line 1184: We should also remove this one right? Notification should be same for both

yarn.lock
```diff
@@ -13316,13 +13321,13 @@ normalize-path@^3.0.0, normalize-path@~3.0.0:
   resolved "https://registry.yarnpkg.com/normalize-path/-/normalize-path-3.0.0.tgz#0dcd69ff23a1c9b11fd0978316644a0388216a65"
   integrity sha512-6eZs5Ls3WtCisHWp9S2GUy8dqkpGi4BVSz3GaqiE6ezub0512ESztXUwUB6C6IKbQkY2Pnb/mD4WYojCRwcwLA==
 
-notistack@^2.0.4:
-  version "2.0.8"
-  resolved "https://registry.yarnpkg.com/notistack/-/notistack-2.0.8.tgz#78cdf34c64e311bf1d1d71c2123396bcdea5e95b"
-  integrity sha512-/IY14wkFp5qjPgKNvAdfL5Jp6q90+MjgKTPh4c81r/lW70KeuX6b9pE/4f8L4FG31cNudbN9siiFS5ql1aSLRw==
+notistack@^3.0.1:
```
FiboApe, line 13324: seems like there was a breaking change, are we handling it correctly?

apps/root/src/pages/aggregator/swap-container/components/swap/index.tsx
th0rOdinson: I wasn't sure how to handle this case. now we open the modal and when it closes, we send the user back to first step.
I wouldn't to the same as this "Better quote" notification, the user would be blocked. We could close the txSteps and render an error notification in the snackbar?

FiboApe (approved):

PR #572, opened 2024-02-16
feat(dca): remove pair and swap intervals graphql dependency

FiboApe:


apps/root/src/services/web3Service.ts
```diff
@@ -167,13 +167,7 @@ export default class Web3Service {
       this.labelService
     );
     this.eventService = new EventService(this.providerService, this.accountService);
-    this.pairService = new PairService(
```
th0rOdinson, line 170: 😍

th0rOdinson (approved):

PR #576, opened 2024-02-21
feat: enables DCA on moonbeam

0xged:


apps/root/package.json
```diff
@@ -84,7 +84,7 @@
   },
   "dependencies": {
     "@apollo/client": "^3.3.20",
-    "@arcxmoney/analytics": "^2.0.2",
+    "@arcxmoney/analytics": "^2.0.3",
```
th0rOdinson, line 87: This upgrade should not be necessary right?

0xged: no need really, but it threw an arcx-related error, so I also took the opportunity to update the package :)

th0rOdinson (approved):

PR #577, opened 2024-02-22
feat(ui-library): add token picker component

FiboApe:
Screenshot 2024-02-22 at 17 34 52
Screenshot 2024-02-22 at 17 34 59


packages/ui-library/src/common/index.ts
```diff
@@ -1,3 +1,35 @@
 export { createStyles } from '@mui/material/styles';
 export { useTheme, useMediaQuery } from '@mui/material';
 export * from './notistack';
+
+export const fallbackCopyTextToClipboard = (text: string) => {
```
th0rOdinson, line 5: I'm not a fan of having a function in this file, I believe it should only be for exporting sibling files/directories. Maybe it's time to have a common/utils directorory (utils package coming soon 😄 )

packages/ui-library/src/components/token-picker/index.tsx
```diff
@@ -0,0 +1,419 @@
...
+) => {
+  const { token, balance, allowsYield, isCustomToken } = tokenWithBalance;
+  const balanceUsd = balance?.amountInUSD;
+  const balanceUnits = balance?.amountInUnits;
+
+  const onCopyAddress = (evt: React.MouseEvent<SVGSVGElement, MouseEvent>) => {
+    evt.stopPropagation();
+    if (token) {
+      copyTextToClipboard(token.address);
+    }
```
th0rOdinson, line 139: What do you think of showing a notification as we do on other copyButtons?

packages/ui-library/src/components/token-picker/index.tsx
```diff
@@ -0,0 +1,419 @@
...
+  }
+
+  return (
+    <Modal open={shouldShow} onClose={onClose} closeOnBackdrop maxWidth="sm" keepMounted fullHeight>
+      <ContainerBox alignSelf="stretch" flex={1}>
+        <IconButton
+          aria-label="close"
+          size="small"
+          onClick={handleOnClose}
+          style={{ position: 'absolute', top: '24px', right: '32px' }}
```
th0rOdinson, line 380: Let's apply theme untis :D

th0rOdinson (approved):
OMFG this is beautiful

PR #578, opened 2024-02-23
feat(history): adds history search

FiboApe:


apps/root/src/common/utils/transaction-history/search.ts
```diff
@@ -0,0 +1,75 @@
...
+  } = event.tx;
+
+  return (
+    name.toLowerCase().includes(search) ||
+    txHash.toLowerCase().includes(search) ||
+    initiatedBy.toLowerCase().includes(search)
+  );
+};
+
+const searchByType = (event: TransactionEvent, search: string) => event.type.toLowerCase().includes(search);
```
th0rOdinson, line 19: We should search by TRANSACTION_TYPE_TITLE_MAP and not by event Type.
I think this is the right way, because in the UI we are showing 'Create DCA Position', but the search is targeting for types included in TransactionEventTypes (would find my DCa create by the string 'DCA Deposted', which is unknown by the user. In this way, the search value would match operation column

FiboApe: oh good catch

th0rOdinson (approved):

PR #579, opened 2024-02-23
feat(swap): Apply styles to advanced settings

th0rOdinson:
### Default
Screenshot 2024-02-23 at 17 30 50

### Collapsed
Screenshot 2024-02-23 at 17 30 57


apps/root/src/pages/aggregator/swap-container/components/swap-settings/components/slippage-input/index.tsx
```diff
@@ -2,72 +2,24 @@ import React from 'react';
...
-  border-color: ${$isSelected ? colors[mode].violet.violet200 : baseColors.disabledText} !important;
-  `}
-`;
-
-const StyledTabContainer = styled.div`
-  display: flex;
-  gap: 8px;
-  margin-left: 8px;
+const StyledTextField = styled(TextField)`
+  width: ${({ theme }) => theme.spacing(37)};
```
FiboApe, line 18: is this necessary? Widths on px are extremely discouraged

packages/ui-library/src/theme/variants/toggle-button-group-variants.ts
```diff
@@ -0,0 +1,39 @@
...
+    },
+  },
+  MuiToggleButton: {
+    styleOverrides: {
+      root: {
+        minWidth: SPACING(24),
+        padding: `${SPACING(2.5)} ${SPACING(3)}`,
+        textTransform: 'none',
+        borderRadius: `${SPACING(2)} !important`,
+        border: `1px solid ${colors[mode].border.border1} !important`,
```
FiboApe, line 21: I love the smell of `!important` in the morning

packages/ui-library/src/components/options-menu/index.tsx
```diff
@@ -115,6 +122,14 @@ const OptionsMenu = ({
           vertical: 'top',
           horizontal: 'left',
         }}
+        slotProps={{
+          paper: {
+            style: {
+              maxHeight: spacing(80),
```
FiboApe, line 128: heyo, why the maxHeight set on px?

packages/ui-library/src/components/background-paper/index.tsx
```diff
@@ -12,7 +12,7 @@ const StyledBackgroundPaper = styled(Paper)`
   }) => `
     background-color: ${colors[mode].background.quartery};
     border-radius: ${spacing(4)};
-    padding: ${spacing(8)};
+    padding: ${spacing(10)} ${spacing(8)};
```
FiboApe, line 15: this doesnt break anything?

apps/root/src/state/aggregator-settings/reducer.ts
```diff
@@ -54,9 +53,6 @@ export default createReducer(initialState, (builder) => {
     .addCase(setShowTransactionCost, (state, { payload }) => {
       state.showTransactionCost = payload;
     })
-    .addCase(setConfetti, (state, { payload }) => {
```
FiboApe, line 57: Saddies :(

apps/root/src/pages/aggregator/swap-container/components/swap-settings/index.tsx
```diff
@@ -107,146 +184,179 @@ const SwapSettings = ({ shouldShow, onClose }: SwapSettingsProps) => {
...
+  );
 
   return (
-    <Slide direction="up" in={shouldShow} mountOnEnter unmountOnExit>
+    <Slide
+      direction="up"
+      in={shouldShow}
+      mountOnEnter
+      unmountOnExit
+      onExited={() => setShouldShowFirstStep(true)}
```
FiboApe, line 222: why is this needed?

packages/ui-library/src/components/background-paper/index.tsx
th0rOdinson: The designer has changed the paddings and margins of some papers and modals in the UI. I created a task for this (BLY-1668), but I needed this change here to have the component finished with this task

apps/root/src/pages/aggregator/swap-container/components/swap-settings/index.tsx
th0rOdinson: This is the same approach I did on transaction steps. The height for swap-settings is different from the height of swap first step. So having the <Slide /> as absolute with overflow: hidden in the container would not work anymore. With this implementation, the First step unmounts when the settings Slide started entering, and mounts when the Slide has exited

packages/ui-library/src/theme/variants/toggle-button-group-variants.ts
```diff
@@ -14,19 +14,20 @@ export const buildToggleButtonGroupVariant = (mode: 'light' | 'dark'): Component
...
-          background: colors[mode].background.tertiary,
+        '&&&.MuiToggleButton-root': {
+          border: `1px solid ${colors[mode].border.border1}`,
+          borderRadius: `${SPACING(2)}`,
+          ':hover': {
+            background: colors[mode].background.tertiary,
+          },
         },
-        '&.Mui-selected': {
+        '&&&.Mui-selected': {
```
FiboApe, line 28: those are a lot of &, why do you need those? should only be one

FiboApe: There's probably a selected style override you can set this there

th0rOdinson: About the specificity, the selector used by MUI is something like `.ktEYlR .MuiToggleButtonGroup-grouped:not(:last-of-type) {`. Having one & isn't being enought. I have downgrade this to && though 😄,

The `selected` is not working, in fact throws an error in the console when using it

th0rOdinson: In order to override border+borderRadius properties, I needed a more specific selector. The one used by MUI is something like `.ktEYlR .MuiToggleButtonGroup-grouped:not(:last-of-type) {`. Having one `&` isn't being enought, so to override this I needed to use `&&&`.
Then, this implied that for `.Mui-selected` case I needed to use a selector with same specificity.

The selected rule is not working, in fact throws an error in the console when using it, suggesting to use .Mui-selected

FiboApe: Right, take a look into the Css classes part of https://mui.com/material-ui/api/button-group/

Just like we target `root`, you can target `contained` and etc

th0rOdinson: I couldn't find a working prop for this case, but I was able to improve this specificity issue

FiboApe (approved):

PR #580, opened 2024-02-24
Bly 1647

FiboApe:


packages/ui-library/src/components/transaction-confirmation/index.tsx
```diff
@@ -0,0 +1,374 @@
...
+            key={action.label}
+            color={action.color}
+            onClick={action.onAction}
+            fullWidth
+            size="large"
+          >
+            {action.label}
+          </Button>
+        ))}
+        <Button variant="outlined" color="primary" fullWidth onClick={onGoToEtherscan} size="large">
```
th0rOdinson, line 237: how about showing the actual receipt component here?

FiboApe: Oh good idea

th0rOdinson (approved):
Great job sir

PR #582, opened 2024-02-27
feat(dca): Modify DCA amount and frequency inputs

th0rOdinson:
Screenshot 2024-02-27 at 15 30 41


apps/root/src/pages/dca/create-position/components/step1/components/frequency-selector/index.tsx
```diff
@@ -47,29 +30,63 @@ const FrecuencySelector = ({ frequencies, handleFrequencyChange }: Props) => {
...
-              ),
-            }}
-          />
-        </Typography>
-        <FrequencyInput id="frequency-value" value={frequencyValue} onChange={handleFrequencyChange} />
-      </StyledFrequencyValueContainer>
-    </StyledFrequencyContainer>
+              )}`}
+              onChange={(evt) => validator(evt.target.value.replace(/,/g, '.'))}
+              disabled={!from}
```
FiboApe, line 66: why this?

FiboApe (approved):

apps/root/src/pages/dca/create-position/components/step1/components/frequency-selector/index.tsx
th0rOdinson: The onChange for this input has a if (!from) return. I wanted to keep this as it was! Are you suggesting to enable this input no matter the rest of the state?

PR #587, opened 2024-02-29
feat(home): No wallet connected case

th0rOdinson:
Fixed:
- No wallet connected on `<NetWorth />` and `<Portfolio />`
- Re-order styles on ContainerBox to enable Skeletons in `<Portfolio />` render as they should
- Elevation for our Virtualized Tables in order to match design 
- Set border-radius value for ForegroundPaper as default, based on main components (such as `Activity`)

### Previous behaviour
Screenshot 2024-02-29 at 17 25 46

### Current Demo
Screenshot 2024-02-29 at 17 52 34
Screenshot 2024-02-29 at 17 51 49


FiboApe (approved):
Niiiiiiiiiiice!!!

PR #590, opened 2024-03-01
feat(aggregator): use different timeouts for rootstock

FiboApe:


apps/root/src/services/sdkService.ts
```diff
@@ -200,6 +200,16 @@ export default class SdkService {
         ...(disabledDexes ? { filters: { excludeSources: disabledDexes } } : {}),
       };
 
+      console.log('taker address', {
```
th0rOdinson, line 203: 😮

th0rOdinson (approved):

PR #591, opened 2024-03-04
feat(*): Add timeout and error handling for initial requests

th0rOdinson:


apps/root/src/state/balances/actions.ts
```diff
@@ -76,37 +77,41 @@ export const fetchInitialBalances = createAppAsyncThunk<
...
-          )
-        );
-
-        set(
-          parsedAccountBalances,
-          [chainId, 'balancesAndPrices', tokenAddress, 'balances', walletAddress],
-          BigInt(balance)
-        );
-        set(parsedAccountBalances, [chainId, 'balancesAndPrices', tokenAddress, 'token'], token);
+  try {
```
FiboApe, line 82: No need to wrap this, since the actions already provide a `.rejected` value just like we use the `success` and `pending`

apps/root/src/frame/components/promises-initializer/index.tsx
```diff
@@ -0,0 +1,109 @@
...
+              </Button>
+            </>
+          ),
+          key: errorKey,
+          preventDuplicate: true,
+          anchorOrigin: {
+            vertical: 'bottom',
+            horizontal: 'right',
+          },
+          persist: true,
```
FiboApe, line 72: I'm not sure we want to persist this right? or maybe we do, probs should check with designer

apps/root/src/frame/components/promises-initializer/index.tsx
```diff
@@ -0,0 +1,109 @@
...
+    },
+    [snackbar, navigate]
+  );
+
+  React.useEffect(() => {
+    const fetchBalancesAndPrices = async () => {
+      // Fire-and-Forget Promises
+      timeoutPromise(contactListService.initializeAliasesAndContacts(), TimeoutPromises.COMMON, {
+        description: ApiErrorKeys.LABELS_CONTACT_LIST,
+      }).catch((error) => handleError(error));
```
FiboApe, line 85: could be just `catch(handleError)`

apps/root/src/frame/components/promises-initializer/index.tsx
```diff
@@ -0,0 +1,109 @@
...
+          anchorOrigin: {
+            vertical: 'bottom',
+            horizontal: 'right',
+          },
+          persist: true,
+          TransitionComponent: Zoom,
+        });
+      }
+    },
+    [snackbar, navigate]
```
FiboApe, line 77: we'll probably also need the intl dependency since I think it changes based on the language

FiboApe (approved):

PR #592, opened 2024-03-04
feat(aggregator): style inputs

FiboApe:


apps/root/src/pages/aggregator/swap-container/components/quote-picker/index.tsx
```diff
@@ -153,19 +153,21 @@ const QuotePicker = ({ quotes, isLoading, bestQuote, isBuyOrder }: QuotePickerPr
...
-            />
-          ),
-        }}
+        token={
+          (selectedRoute && {
+            ...toToken({ symbol: selectedRoute?.swapper.name }),
+            icon: (
+              <TokenIcon
+                isInChip
+                token={!isLoading ? emptyTokenWithLogoURI(selectedRoute?.swapper.logoURI || '') : undefined}
```
th0rOdinson, line 165: We can remove '?' after both selectedRoute

packages/ui-library/src/components/token-picker-button/index.tsx
```diff
@@ -36,30 +50,28 @@ const StyledKeyboardArrowDownIcon = styled(KeyboardArrowDownIcon)`
...
   isLoading,
   onClick,
   showAction = true,
   defaultText,
   disabled,
 }: TokenPickerButtonProps) => (
   <StyledTokenPickerButton disabled={disabled} onClick={onClick}>
-    {token.icon}
-    <StyledActiveLabel $isSelected={isSelected}>
+    {token?.icon || <StyledEmptyTokenIcon $realSize={SPACING(7)} />}
```
th0rOdinson, line 70: I dont find the emptyTokenIcon size flexible enough. We could do one of two:
1. Always expect a token. If the consumer of this component doesn't have one, it should do 
`token = {...toToken({icon: <TokenIcon size={X} />})}`. The empty token in TokenIcon can be this StyledEmptyTokenIcon
2. Get the realSize as a param
What do you think?

apps/root/src/pages/aggregator/swap-container/components/step1/index.tsx
```diff
@@ -84,29 +93,41 @@ const SwapFirstStep = ({
       : fromValue;
   let toValueToUse = isBuyOrder
     ? toValue
-    : (selectedRoute?.buyToken.address === to?.address &&
-        formatUnits(BigInt(selectedRoute?.buyAmount.amount || '0'), selectedRoute?.buyToken.decimals || 18)) ||
+    : (selectedRoute &&
+        selectedRoute?.buyToken.address === to?.address &&
+        formatUnits(selectedRoute?.buyAmount.amount, selectedRoute.buyToken.decimals || 18)) ||
```
th0rOdinson, line 98: Can remove '?' also here

apps/root/src/common/components/token-amount-input/index.tsx
```diff
@@ -65,50 +55,121 @@ const TokenAmountInput = ({
...
+                autoComplete="off"
+                placeholder="0"
+                disableUnderline
+                inputProps={{ style: { textAlign: 'right', height: 'auto' } }}
+                sx={{ ...buildTypographyVariant(mode).h3, fontWeight: '700', color: 'inherit', textAlign: 'right' }}
+              />
+            </FormControl>
+            <ContainerBox>
+              <Typography variant="body">≈{` $${tokenAmount.amountInUSD || '0'}`}</Typography>
+              {priceImpact && !isNaN(Number(priceImpact)) && isFinite(Number(priceImpact)) && (
```
th0rOdinson, line 152: We should add a gap={1} between priceImpact and amountInUSD. Also the color for priceImpact doesn't adjust to light theme

Screenshot 2024-03-04 at 17 16 15

th0rOdinson (commented):
This looks great 😄 !

apps/root/src/common/components/token-amount-input/index.tsx
```diff
@@ -147,17 +147,17 @@ const TokenAmountInput = ({
...
                   variant="body"
                   color={
                     // eslint-disable-next-line no-nested-ternary
                     Number(priceImpact) < -2.5
-                      ? colors[mode].semantic.error.primary
+                      ? colors[mode].semantic.error.darker
                       : Number(priceImpact) > 0
-                      ? colors[mode].semantic.success.primary
+                      ? colors[mode].semantic.success.darker
                       : baseColors.disabledText
```
th0rOdinson, line 161: I believe disabledText should be changed to a theme-based color
Screenshot 2024-03-04 at 18 09 11

th0rOdinson (approved):

apps/root/src/common/components/token-amount-input/index.tsx
FiboApe: oh good point

PR #599, opened 2024-03-06
feat(dca): add provitional faq and landing

FiboApe:


apps/root/src/pages/dca/create-position/components/landing/chart.tsx
```diff
@@ -0,0 +1,132 @@
...
+        fill="#978AA8"
+      />
+      <line x1="52" y1="108.741" x2="577" y2="108.741" stroke="#A399FF" strokeWidth="3" strokeDasharray="15 15" />
```
th0rOdinson, line 25: This should be violet600 (themeMode-based)

apps/root/src/pages/dca/create-position/components/landing/chart.tsx
```diff
@@ -0,0 +1,132 @@
...
+      <defs>
+        <linearGradient
+          id="paint0_linear_6629_32270"
+          x1="595.399"
+          y1="88.6885"
+          x2="595.399"
+          y2="265.349"
+          gradientUnits="userSpaceOnUse"
+        >
+          <stop stopColor="#791AFF" />
```
th0rOdinson, line 123: We can also replace this with violet500, it's the same in both modes, but it doesn't hurt to handle it this way 😄

apps/root/src/pages/dca/create-position/components/landing/index.tsx
```diff
@@ -0,0 +1,85 @@
...
+  const intl = useIntl();
+
+  const logoProps = { size: spacing(16), fill: colors[palette.mode].typography.typo2 };
+  const logo =
+    palette.mode === 'light' ? <BalmyLogoSmallDark {...logoProps} /> : <BalmyLogoSmallLight {...logoProps} />;
+
+  return (
+    <StyledBackgroundPaper variant="outlined">
+      {logo}
+      <Typography variant="h4" fontWeight={700}>
```
th0rOdinson, line 64: It looks better with textAlign center imho

apps/root/src/pages/dca/frame/index.tsx
```diff
@@ -68,14 +67,14 @@ const DcaFrame = ({ isLoading }: DcaFrameProps) => {
           <ContainerBox flexDirection="column" gap={8}>
             <NetWorth walletSelector={{ options: { setSelectionAsActive: true } }} />
             {currentRoute === DCA_CREATE_ROUTE.key ? (
-              <CreatePosition handleChangeNetwork={handleChangeNetwork} />
+              <>
+                <CreatePosition handleChangeNetwork={handleChangeNetwork} />
+              </>
```
th0rOdinson, line 72: No need for fragment here

apps/root/src/pages/dca/create-position/components/landing/chart.tsx
FiboApe: You sure? I exported the svg directly from figma and it exported the same one both for dark and light

apps/root/src/pages/dca/create-position/components/landing/chart.tsx
FiboApe: Same as before

apps/root/src/pages/dca/create-position/components/landing/chart.tsx
th0rOdinson: Maybe i'm too meticulous at this point 😅 But this is the dashed line, and in light mode is looking much brighter comparing it with the design. I find this particular vector to change based on the theme

th0rOdinson (approved):

PR #601, opened 2024-03-06
feat(theme): update theme variables and button variants

FiboApe:


packages/ui-library/src/theme/variants/button-variants.ts
```diff
@@ -1,112 +1,62 @@
...
-          ...buildButtonDisabledVariants(mode, ownerState),
-          // common attributes
-          textTransform: 'none',
-          padding: `${SPACING(3)} ${SPACING(6)}`,
-        };
+      // root: ({ ownerState }) => {
+      root: {
+        // return {
+        //   ...buildButtonDisabledVariants(mode, ownerState),
+        // common attributes
```
th0rOdinson, line 12: We could get rid of this comments

packages/ui-library/src/theme/variants/button-variants.ts
```diff
@@ -1,112 +1,62 @@
...
+        textTransform: 'none',
+        padding: `${SPACING(3)} ${SPACING(6)}`,
+        // };
       },
       endIcon: {
         marginLeft: `${DEFAULT_SPACING}px`,
       },
-      outlined: {
-        border: `1px solid ${colors[mode].border.border1}`,
+      outlinedPrimary: {
```
th0rOdinson, line 20: Oh this is much better than a secondary button that is outlined 😄 . Nice job!

apps/root/src/frame/components/background-grid/light.tsx
```diff
@@ -0,0 +1,753 @@
+import React from 'react';
+import { CustomSvgIcon, SvgIconProps } from 'ui-library';
+
+interface IconProps extends SvgIconProps {
+  size?: string;
+}
+
+export default function LightBackgroundGrid({ size, ...props }: IconProps) {
+  return (
+    <CustomSvgIcon viewBox="0 0 1921 668" {...props} sx={{ width: '100%', height: '100%' }} fill="none">
```
th0rOdinson, line 10: I believe we should split the Grid from the gradient in different approaches. The grid is well positioned, but the gradient should continue bellow the 100vh
Screenshot 2024-03-06 at 14 35 54

I did something similar in the Swap landing!

FiboApe: Oh shit forgot to set it as fixed

th0rOdinson: It's not reaching the bottom yet 😢 
Screenshot 2024-03-06 at 15 27 07

th0rOdinson (approved):

PR #612, opened 2024-03-11
Bly 1768

FiboApe:


apps/root/src/frame/index.tsx
```diff
@@ -180,7 +179,6 @@ const AppFrame = ({ config: { wagmiClient, chains }, initialChain }: AppFramePro
                               path="/transfer/:chainId?/:token?/:recipient?"
                               element={<Transfer isLoading={isLoadingNetwork} />}
                             />
-                            <Route path="/euler-claim" element={<EulerClaimFrame isLoading={isLoadingNetwork} />} />
```
th0rOdinson, line 183: ✨

apps/root/src/pages/dca/positions/components/usd-dashboard/index.tsx
```diff
@@ -425,212 +270,107 @@ const UsdDashboard = ({ selectedChain, onSelectTokens, selectedTokens }: UsdDash
...
-                    <Typography variant="bodySmall">
-                      {!positionCountLabel.isOther
-                        ? formatCurrencyAmount(positionCountLabel.summedBalanceToShow, positionCountLabel.token, 4)
-                        : `$${positionCountLabel.summedBalanceUsdToShow.toFixed(2)}`}
-                    </Typography>
-                  </Grid>
+                  )}
+                </Grid>
+                <Grid item xs={3} sx={{ textAlign: 'right' }}>
+                  <Typography variant="bodySmall">{`$${positionCountLabel.value.toFixed(2)}`}</Typography>
```
th0rOdinson, line 366: template literal is not necessary. Also, could textAlign be applied directly to the Typography component?

apps/root/src/services/positionService.ts
```diff
@@ -1500,6 +1544,7 @@ export default class PositionService {
     }
   }
 
+  // TODO: Add price and amounts in usd
```
th0rOdinson, line 1547: Is this still in TODO? I'm not finding where this is pending

th0rOdinson:
I think we should disable click events for the pie chart:
Screenshot 2024-03-11 at 18 49 23

apps/root/src/pages/dca/positions/components/usd-dashboard/index.tsx
FiboApe: I'll remove the template literal but the text align has to stay since typography is inline-block and cannot use the textAlign

apps/root/src/services/positionService.ts
FiboApe: Oh shit good catch, actually added this comments so I wouldn't forget, forgot anyway u.u

packages/ui-library/src/components/dashboard/index.tsx
```diff
@@ -0,0 +1,216 @@
...
+  valuesForOther?: number;
+  valueFormatter?: (value: number) => string;
+}
+
+type DataWithFill = Data & {
+  fill: string;
+  relativeValue: number;
+};
+
+const COLOR_PRIORITIES = {
```
th0rOdinson, line 68: Found something interesting... If you change the theme mode in the application, colors in the charts doesn't change. Only in first render they are defined

apps/root/package.json
```diff
@@ -120,7 +120,7 @@
     "react-virtualized-auto-sizer": "^1.0.5",
     "react-virtuoso": "^4.6.2",
     "react-window": "^1.8.6",
-    "recharts": "^2.1.9",
+    "recharts": "2.8.0",
```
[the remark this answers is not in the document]
FiboApe: We still use it one the root app, but yeah the idea is to move away from that as we keep migrating things to ui library

th0rOdinson (approved):

th0rOdinson (approved):

PR #614, opened 2024-03-13
feat(dca): redirect to create if user has no positions, else show pos…

FiboApe:
…ition list

### DIFF
--- a/apps/root/src/frame/components/navigation/index.tsx
+++ b/apps/root/src/frame/components/navigation/index.tsx
@@ -3,11 +3,10 @@ import ProfileSelector from '@common/components/profile-selector';
 import {
   DASHBOARD_ROUTE,
   DCA_ROUTE,
-  DCA_CREATE_ROUTE,
-  DCA_POSITIONS_ROUTE,
   SWAP_ROUTE,
   TRANSFER_ROUTE,
   HOME_ROUTES,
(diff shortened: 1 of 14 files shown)


apps/root/src/frame/components/navigation/index.tsx
```diff
@@ -137,10 +136,7 @@ const Navigation = ({
           ...DCA_ROUTE,
           label: intl.formatMessage(DCA_ROUTE.label),
           type: SectionType.link,
-          options: [
-            { ...DCA_CREATE_ROUTE, label: intl.formatMessage(DCA_CREATE_ROUTE.label), type: SectionType.link },
-            { ...DCA_POSITIONS_ROUTE, label: intl.formatMessage(DCA_POSITIONS_ROUTE.label), type: SectionType.link },
-          ],
+          activeKeys: [DCA_ROUTE.key, DCA_CREATE_ROUTE.key],
```
th0rOdinson, line 139: Found one case here: If the user clicks on `Invest (DCA)`, the url would change into /positions every time.
This is annoying when you have been redirected to /create, and you click again on `Invest (DCA)`, the url changes but the site doesnt. 
I think we should also change the url of the navigation item if we have fetched this information in positionService

th0rOdinson (approved):

PR #633, opened 2024-03-25
feat(dca): add new styles to graphs

th0rOdinson:


apps/root/src/pages/position-detail/components/position-summary-controls/index.tsx
```diff
@@ -370,58 +369,55 @@ const PositionSummaryControls = ({ pendingTransaction, position }: PositionSumma
...
-  const options = React.useMemo<OptionsMenuOption[]>(
-    () => [
-      {
-        type: OptionsMenuOptionType.option,
-        label: intl.formatMessage(
-          defineMessage({
-            description: 'withdrawToken',
-            defaultMessage: 'Withdraw {token}',
-          }),
+  const options = [
```
th0rOdinson, line 372: This change solves a bug: due to the early return in this function, we had the 'Rendered fewer hooks than expected' error. We could move this array up but I don't think we'd gain too much by memoizing it

apps/root/src/pages/position-detail/components/graph-container/components/swap-price-graph/index.tsx
```diff
@@ -134,35 +87,35 @@ const AveragePriceGraph = ({ position }: AveragePriceGraphProps) => {
           position.chainId,
           swappedActions.length * FREQUENCY_TO_MULTIPLIER[position.swapInterval.toString()],
           FREQUENCY_TO_PERIOD[position.swapInterval.toString()],
-          swappedActions[0].tx.timestamp.toString()
+          swappedActions[swappedActions.length - 1] && swappedActions[swappedActions.length - 1].tx.timestamp.toString()
```
FiboApe, line 90: why this change? was it wrong?

apps/root/src/pages/position-detail/components/position-summary-controls/index.tsx
FiboApe: Mmmm seems like this a solution for a code smell. We would this not be rendered after X action? that makes the error throw?

apps/root/src/pages/position-detail/components/graph-container/components/swap-price-graph/index.tsx
th0rOdinson: [0] is the first timestamp of the position and we needed the last one for this `end` parameter

apps/root/src/pages/position-detail/components/position-summary-controls/index.tsx
th0rOdinson: Kinda, This options are only displayed if the connected wallet is position owner. If the user is not connected and then connects the wallet, being the owner of the position, the component re-renders and has this hooks issue.
It's not recommended anyway to have hooks after an early return based on [official docs](https://arc.net/l/quote/ghusmmuw)

I'm removing the early return so we can keep the memo :)

FiboApe: Yeah and lets fix this so it doesnt render until we have the user and positions

apps/root/src/pages/position-detail/components/position-summary-controls/index.tsx
```diff
@@ -369,162 +367,167 @@ const PositionSummaryControls = ({ pendingTransaction, position }: PositionSumma
...
-              values={{
-                token:
-                  hasSignSupport || position.to.address !== PROTOCOL_TOKEN_ADDRESS
-                    ? position.to.symbol
-                    : wrappedProtocolToken.symbol,
-              }}
-            />
-          </Button>
-        )}
+    ownerWallet && (
```
FiboApe, line 424: We could check this before even rendering the component if we are just going to return nothing if the wallet is not there

FiboApe (approved):

PR #637, opened 2024-03-26
feat(dca): Position card skeleton

th0rOdinson:


apps/root/src/pages/dca/positions/components/positions-list/current-positions/index.tsx
```diff
@@ -227,18 +221,24 @@ const CurrentPositions = ({ isLoading }: CurrentPositionsProps) => {
...
-                  onWithdraw={onWithdraw}
-                  onReusePosition={onShowModifyRateSettings}
-                  onTerminate={onShowTerminate}
-                  disabled={false}
-                  hasSignSupport={!!hasSignSupport}
-                />
-              </StyledGridItem>
-            ))}
+            {isLoading
+              ? Array.from(Array(5).keys()).map((i) => (
```
FiboApe, line 225: can we do the skeletonRows pattern like we have on other components?

FiboApe (approved):

PR #640, opened 2024-03-26
Bly 1593

th0rOdinson:


apps/root/src/pages/history/components/historyTable/index.tsx
```diff
@@ -249,16 +259,16 @@ const getTxEventRowData = (txEvent: TransactionEvent, intl: ReturnType<typeof us
     ) : txDate.startOf('day').equals(DateTime.now().minus({ days: 1 }).startOf('day')) ? (
       <FormattedMessage defaultMessage="Yesterday" description="yesterday" />
     ) : (
-      <>{txDate.toLocaleString(DateTime.DATE_SHORT)}</>
+      <>{txDate.toFormat('LL/dd/yyyy')}</>
```
FiboApe, line 262: can we use toLocaleString for this? Or at least find some DateTime that fixes into this?

The main thing about this is that intl can parse base on localization and show those dates specific for each locale.

For example on use it does mm/dd/yyyy but on latam it does dd/mm/yyyy

packages/ui-library/src/theme/components.ts
```diff
@@ -122,15 +122,25 @@ const baseComponents: Components = {
...
           borderLeftStyle: 'solid',
           borderTopRightRadius: SPACING(4),
           borderBottomRightRadius: SPACING(4),
+          paddingRight: SPACING(8),
           borderColor: 'transparent',
         },
       },
+      head: {
+        ':first-child': {
+          paddingLeft: SPACING(8),
```
FiboApe, line 138: This doesnt break any other table right? what about the portfolio one in the home?

apps/root/src/pages/history/components/historyTable/index.tsx
```diff
@@ -249,16 +271,16 @@ const getTxEventRowData = (txEvent: TransactionEvent, intl: ReturnType<typeof us
     ) : txDate.startOf('day').equals(DateTime.now().minus({ days: 1 }).startOf('day')) ? (
       <FormattedMessage defaultMessage="Yesterday" description="yesterday" />
     ) : (
-      <>{txDate.toLocaleString(DateTime.DATE_SHORT)}</>
+      <>{txDate.toLocaleString({ day: '2-digit', month: '2-digit', year: 'numeric' })}</>
     );
     dateTime = {
       date: formattedDate,
-      time: txDate.toLocaleString(DateTime.TIME_24_WITH_SHORT_OFFSET),
+      time: txDate.toFormat('HH:mm ZZZZ'),
```
FiboApe, line 278: same with this ones for locale strings since for the locale it also uses the timezone of the user to format the time :D

FiboApe (approved):

PR #642, opened 2024-03-27
BLY-2015

th0rOdinson:


yarn.lock
```diff
@@ -2606,7 +2606,7 @@
     "@uniswap/v3-core" "1.0.1"
     moment "2.29.3"
 
-"@mean-finance/sdk@*", "@mean-finance/sdk@0.0.217":
+"@mean-finance/sdk@*":
```
FiboApe, line 2609: This should not be here, add the `*` dependency to the 223 version

FiboApe (approved):

PR #646, opened 2024-03-28
Bly 1844

FiboApe:


apps/root/src/frame/components/navigation/index.tsx
```diff
@@ -83,25 +83,25 @@ const Navigation = ({
   React.useEffect(() => {
     if (HOME_ROUTES.includes(location.pathname)) {
       dispatch(changeRoute('home'));
-      pushToHistory('/');
+      // pushToHistory('/');
```
th0rOdinson, line 86: Should remove this comments, may be confusing in the future 😄

th0rOdinson (approved):

apps/root/src/frame/components/navigation/index.tsx
FiboApe: Oh thanks

PR #648, opened 2024-03-28
fix(transfer): Enable editing contacts

th0rOdinson:


packages/ui-library/src/theme/typography.ts
```diff
@@ -125,7 +125,7 @@ export const buildTypographyVariant = (mode: PaletteMode): TypographyOptions =>
     fontFamily: 'Inter',
     fontSize: '0.875rem', // 14/16
     fontStyle: 'normal',
-    fontWeight: 400,
+    fontWeight: 500,
```
FiboApe, line 128: hold up there cowboy, this will surely break other styles

packages/ui-library/src/theme/typography.ts
```diff
@@ -140,7 +140,7 @@ export const buildTypographyVariant = (mode: PaletteMode): TypographyOptions =>
     fontFamily: 'Inter',
     fontSize: '0.625rem', // 10/16
     fontStyle: 'normal',
-    fontWeight: 400,
+    fontWeight: 500,
```
FiboApe, line 143: same for this

packages/ui-library/src/theme/variants/button-variants.ts
```diff
@@ -12,6 +12,10 @@ export const buildButtonVariant = (mode: 'light' | 'dark'): Components => ({
       endIcon: {
         marginLeft: `${DEFAULT_SPACING}px`,
       },
+      sizeLarge: {
+        fontSize: '1rem',
```
FiboApe, line 16: how did you calculate this?

packages/ui-library/src/theme/variants/input-base-variants.ts
```diff
@@ -42,7 +42,9 @@ export const buildInputBaseVariant = (mode: 'light' | 'dark'): Components => ({
         },
         input: {
           '&::placeholder': {
-            fontWeight: '700',
+            fontWeight: '600',
```
FiboApe, line 45: same for this one, this might break some stuff right?

apps/root/src/pages/dca/positions/components/positions-list/current-positions/components/position-options/index.tsx
```diff
@@ -79,7 +79,7 @@ const PositionOptions = ({
           color="inherit"
           onClick={(e) => e.preventDefault()}
         >
-          <Typography variant="bodySmall">
+          <Typography variant="bodySmall" color="inherit">
```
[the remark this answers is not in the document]
th0rOdinson: bodySmall has a default color of typo3, so it was overriding the needed color from optionsMenu (which is typo2). I have now removed this typography wrapper and looks much simpler

apps/root/src/pages/transfer/components/recipient-address/components/contact-modal/index.tsx
```diff
@@ -0,0 +1,265 @@
...
+})`
+  max-height: ${({ theme }) => theme.spacing(67)};
+  overflow: scroll;
+`;
+
+export const StyledContactModalParagraph = styled(Typography).attrs({
+  variant: 'body',
+  fontWeight: 500,
+  textAlign: 'center',
+  lineHeight: 1,
```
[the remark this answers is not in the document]
th0rOdinson: It shouldn't be, thank you!

apps/root/src/pages/transfer/components/recipient-address/components/contact-modal/index.tsx
```diff
@@ -0,0 +1,265 @@
...
+          />
+        ),
+      },
+      [ContactListActiveModal.EDIT_CONTACT]: {
+        title: <FormattedMessage description="editContactTitle" defaultMessage="Edit your Contact" />,
+        content: editingContact ? <EditContactModal setActiveModal={setActiveModal} contact={editingContact} /> : <></>,
+      },
+      [ContactListActiveModal.NONE]: { content: <></> },
+    }),
+    [contactList, postContactStatus, defaultAddressValue, activeModal]
```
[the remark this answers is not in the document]
th0rOdinson: good catch!! Thanks

packages/ui-library/src/theme/typography.ts
th0rOdinson: Haha 😄 . I checked this in detail, I truly think this changes (all 3 font weight related) are aligned with the design

packages/ui-library/src/theme/variants/button-variants.ts
th0rOdinson: Just saw in the design this should be button text styles for large case

packages/ui-library/src/theme/typography.ts
FiboApe: Apart from that, what other part of the app that used this is now not according to the design?

FiboApe (approved):
Go for it

PR #650, opened 2024-03-29
feat(*): add feedback tracking for DCA and Aggregator

FiboApe:


packages/ui-library/src/components/thumbs-satisfaction/index.tsx
```diff
@@ -0,0 +1,88 @@
...
+      $selected &&
+      `
+    background-color: ${colors[palette.mode].background.tertiary};
+    outline-color ${colors[palette.mode].border.accent};
+    `
+    }
+`}
+`;
+
+const ThumbsSatisfaction = ({ id, onClickOption }: ThumbsSatisfactionProps) => {
```
th0rOdinson, line 36: Why not using customer-satisfaction component directly?

FiboApe: It has the labels on the top, values are numbers and have to parse them. Just overall simplified streamed experience using this one, plus some styles are also differente like the text being left aligned instead of center aligned etc

th0rOdinson (approved):

PR #652, opened 2024-03-31
fix(datetime): removes timezone

0xged:


apps/root/src/pages/history/components/historyTable/index.tsx
```diff
@@ -270,18 +270,23 @@ const getTxEventRowData = (txEvent: TransactionEvent, intl: ReturnType<typeof us
...
+      <>{txDate.toLocaleString({ day: '2-digit', month: '2-digit', year: 'numeric', timeZoneName: undefined })}</>
     );
     dateTime = {
       date: formattedDate,
-      time: txDate.toLocaleString({ ...DateTime.TIME_24_WITH_SHORT_OFFSET, second: undefined }),
+      time: txDate.toLocaleString({
+        ...DateTime.TIME_24_WITH_SHORT_OFFSET,
+        second: undefined,
+        timeZoneName: undefined,
+      }),
```
th0rOdinson, line 281: Here it should be enough to use DateTime.TIME_24_SIMPLE (Same for line 289)

th0rOdinson (approved):

PR #653, opened 2024-03-31
fix(composed-token-icon): token and network icon borders

0xged:
It also fixes the "assets" icons that had the same issue. No icon uses borders in general on the design, so I guess this will not break anything.


th0rOdinson (approved):
Looks great!

PR #654, opened 2024-03-31
fix(portfolio): skeleton missing, column order and thickness

0xged:
While doing BLY-1980, I also detected that:

- There was a loading skeleton missing
- Thickness could look more similar to Figma's
- Also, if the screen loaded slow enough (set network to throttle) the percentage could show 'NaN' as value, added that to the logic of showing / not showing the value.

Prev
image

New
image

### DIFF
--- a/apps/root/src/pages/home/components/portfolio/index.tsx
+++ b/apps/root/src/pages/home/components/portfolio/index.tsx
@@ -134,7 +134,6 @@ const PortfolioNotConnected = () => {
     </StyledNoWallet>
   );
 };
-
 const PortfolioBodyItem: ItemContent<BalanceItem, Record<string, never>> = (
   index: number,
   { balance, token, isLoadingPrice, price, balanceUsd, relativeBalance }: BalanceItem
@@ -152,9 +151,7 @@ const PortfolioBodyItem: ItemContent<BalanceItem, Record<string, never>> = (
       </TableCell>
       <TableCell>
         <ContainerBox flexDirection="column">
-          <StyledBodyTypography>
-            {formatCurrencyAmount(balance, token, 3)} {token.symbol}
-          </StyledBodyTypography>
+          <StyledBodyTypography>{formatCurrencyAmount(balance, token, 3)}</StyledBodyTypography>
           <StyledBodySmallTypography>
             {isLoadingPrice && !price ? (
               <Skeleton variant="text" animation="wave" />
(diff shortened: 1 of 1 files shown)


apps/root/src/pages/home/components/portfolio/index.tsx
```diff
@@ -170,11 +172,11 @@ const PortfolioBodyItem: ItemContent<BalanceItem, Record<string, never>> = (
       </TableCell>
       {relativeBalance !== 0 && (
         <TableCell>
-          {isLoadingPrice && !price ? (
+          {isNaN(relativeBalance) || isLoadingPrice || !price ? (
```
th0rOdinson, line 175: I think we should catch this NaN result earlier in order to prevent it, otherwise we may show a skeleton even if there is an operation error
I'm sure this is happening because of a 0/0 operation

FiboApe: agree

0xged: The issue comes from L260 where `value.balanceUsd` is undefined, therefore it's  0 divided by `assetsTotalValue.wallet` which is also zero. Seems that this math shouldn't be done while everything is still zero. `isLoadingBalances` is false at the same time. If the new push doesn't satisfy the review, I'd just scrap it and create a linear ticket. My idea was mainly to fix visual stuff, because clearly scope will be missing a lot on these type of fixes.

0xged: @FiboApe @th0rOdinson  wdyt?

apps/root/src/pages/home/components/portfolio/index.tsx
```diff
@@ -255,7 +257,7 @@ const Portfolio = ({ selectedWalletOption }: PortfolioProps) => {
     const mappedBalances = map(Object.entries(tokenBalances), ([key, value]) => ({
       ...value,
       key,
-      relativeBalance: ((value.balanceUsd || 0) / assetsTotalValue.wallet) * 100,
+      relativeBalance: ((value.balanceUsd || 0) / assetsTotalValue.wallet) * 100 || 0,
```
th0rOdinson, line 260: I would avoid 0/0 division, so it would be okay with a ternary to check assetsTotalValue.wallet.
`assetsTotalValue.wallet && value.balanceUsd ? ((value.balanceUsd / assetsTotalValue.wallet) * 100) : 0`

apps/root/src/pages/home/components/portfolio/index.tsx
th0rOdinson: Left one comment related 😄

apps/root/src/pages/home/components/portfolio/index.tsx
0xged: @th0rOdinson sounds about right. done.

th0rOdinson (approved):

PR #655, opened 2024-03-31
fix(navigation): translation and audits icons, filters selected network

0xged:
image
image

=> 

image
image


apps/root/src/frame/components/navigation/index.tsx
```diff
@@ -38,6 +37,7 @@ import { useThemeMode } from '@state/config/hooks';
 import useSelectedLanguage from '@hooks/useSelectedLanguage';
 import { SUPPORTED_LANGUAGES_STRING, SupportedLanguages } from '@constants/lang';
 import useChangeLanguage from '@hooks/useChangeLanguage';
+import { LangIcon } from 'ui-library/src';
```
FiboApe, line 40: This import is incorrect, it should be on the top

packages/ui-library/src/icons/lang.tsx
```diff
@@ -0,0 +1,23 @@
+import React from 'react';
+import { CustomSvgIcon, SvgIconProps } from '../components/svgicon';
+
+interface IconProps extends SvgIconProps {
+  size?: string;
+}
+
+export default function LangIcon({ size, ...props }: IconProps) {
+  return (
+    <CustomSvgIcon viewBox="0 0 25 25" style={size ? { fontSize: size } : {}} {...props}>
```
FiboApe, line 10: You've checked that the viewbox matches the one from your initially exported svg right?

apps/root/src/frame/components/navigation/index.tsx
0xged: Yikes, autoimport why u do dis. Ty.

packages/ui-library/src/icons/lang.tsx
0xged: out of sheer luck, it did 🥶

PR #657, opened 2024-04-02
fix(agg): Prevent simulation with incorrect network

th0rOdinson:


apps/root/src/services/aggregatorService.ts
```diff
@@ -108,19 +108,13 @@ export default class AggregatorService {
     sourceTimeout = TimeoutKey.patient
   ) {
     const currentNetwork = await this.providerService.getNetwork();
+    const balance = await this.walletService.getBalance(from.address, takerAddress);
 
     const isOnNetwork = !chainId || currentNetwork.chainId === chainId;
-    let shouldValidate = !buyAmount && isOnNetwork;
```
0xged, line 113: We are removing the !buyAmount validation btw.

apps/root/src/services/aggregatorService.ts
```diff
@@ -183,7 +177,7 @@ export default class AggregatorService {
       hasEnoughForSwap = sortedOptions.some((option) => balance.gte(option.sellAmount.amount));
     }
 
-    if (usePermit2 && from.address === protocolToken.address && takerAddress && hasEnoughForSwap) {
+    if (usePermit2 && from.address === protocolToken.address && takerAddress && hasEnoughForSwap && isOnNetwork) {
```
0xged, line 180: That is a lot of validation 😂

0xged (approved):

PR #658, opened 2024-04-02
rebranding-hardening

th0rOdinson:


0xged (approved):

apps/root/src/pages/home/components/portfolio/index.tsx
```diff
@@ -153,7 +153,7 @@ const PortfolioBodyItem: ItemContent<BalanceItem, Record<string, never>> = (
       <TableCell>
         <Grid container flexDirection={'row'} alignItems={'center'} gap={3}>
           <TokenIconWithNetwork token={token} />
-          <ContainerBox flexDirection="column">
+          <ContainerBox flexDirection="column" flex="1" style={{ overflow: 'hidden' }}>
```
FiboApe, line 156: this doesnt actually truncate them correctly, can you use an ellipsis for this? with `text-overflow` `ellipsis`

th0rOdinson: The text-overflow ellipsis was already applied because of StyledBodySmallTypography's noWrap prop
The thing is this text isnt technically overflowing , their parent ContainerBox is. So giving flex 1 adapted it's height from parent's availbable space, and overflow-hidden finished the job

Screenshot 2024-04-04 at 3 45 56 PM

FiboApe (approved):

PR #661, opened 2024-04-04
fix(agg): Site crashes after swap

th0rOdinson:


apps/root/src/services/aggregatorService.ts
```diff
@@ -380,16 +380,20 @@ export default class AggregatorService {
...
+          if (parsedLog.eventName === eventName) {
             // @ts-expect-error typescript magic
-            result.push(parsedLog);
+            if (!extraFilter || extraFilter(parsedLog)) {
+              // @ts-expect-error typescript magic
+              result.push(parsedLog);
+            }
           }
+        } catch {
+          // Skipping any event not included in our ABI
```
FiboApe, line 396: we could 'console.error' the message just to debug it in instances where it fails and it shouldnt have

FiboApe (approved):

PR #666, opened 2024-04-06
Bly 1990

FiboApe:


packages/ui-library/src/theme/typography.ts
```diff
@@ -143,4 +143,12 @@ export const buildTypographyVariant = (mode: PaletteMode): TypographyOptions =>
     fontWeight: 500,
     lineHeight: 1.2, // 12/10
   },
+  confirmationLoading: {
```
th0rOdinson, line 146: Not a fan of this, we have just one use case for this. What do you think of adding this as a StyledTypography?

th0rOdinson (approved):

PR #672, opened 2024-04-08
fix(transfer): improve select render item logic

FiboApe:


0xged:
I won't review it because not an FE boi, but thanks for re-taking this issue Fibo <3

PR #675, opened 2024-04-08
feat(agg): swap and transfer with contact list

th0rOdinson:


FiboApe:
If I click a contact and close the modal but not hit "confirm address". It still sets it as the address to transfer the funds. We should probably only set the transferTo on the "confirm address" or close the modal if I'm selecting a contact.

Swap and transfer button should also be disabled while we search for quotes

Also would be good to have on the recap data that we are sending the tokens to another address

FiboApe (approved):

FiboApe (approved):

PR #677, opened 2024-04-08
fix(modify-position): improve modify positions styles

FiboApe:


packages/ui-library/src/components/divider/index.tsx
```diff
@@ -1,4 +1,36 @@
...
+  ${({
+    theme: {
+      palette: { mode },
+    },
+  }) => `
+    border-color: ${colors[mode].border.border1};
+  `}
+`;
+
+const DividerBorder2 = styled(Divider)`
```
th0rOdinson, line 26: Like it very much

packages/ui-library/src/components/modal/index.tsx
```diff
@@ -182,14 +167,19 @@ const Modal: React.FC<ModalProps> = ({
...
                 key={index}
+                sx={{ width: '100%' }}
               >
                 {action.label}
               </Button>
             )
           )}
-        </StyledDialogActions>
+          {showCloseButton && (
+            <Button onClick={onClose} variant="outlined" size="large" sx={{ width: '100%' }}>
```
th0rOdinson, line 178: Doesn't native fullWidth work in thi case? Same for L171

packages/ui-library/src/theme/typography.ts
```diff
@@ -104,7 +104,7 @@ export const buildTypographyVariant = (mode: PaletteMode): TypographyOptions =>
     fontSize: '1rem', // 16/16
     fontStyle: 'normal',
     fontWeight: 400,
-    lineHeight: 'normal', // 32/16
+    lineHeight: 2, // 32/16
```
th0rOdinson, line 107: I believe this is true, but only when body is bold

packages/ui-library/src/components/modal/index.tsx
FiboApe: It adds a few more styles that we dont want

packages/ui-library/src/theme/typography.ts
FiboApe: Oh let me check that

th0rOdinson (approved):

PR #679, opened 2024-04-09
feat(*): Handling non-indexed transactions

th0rOdinson:


apps/root/src/common/utils/transaction-history/parsing.tsx
```diff
@@ -592,4 +596,336 @@ const parseMultipleTransactionApiEventsToTransactionEvents = (
...
+        chainId: event.chainId,
+        txHash: event.hash as Address,
+        timestamp: event.addedTime,
+        explorerLink: buildEtherscanTransaction(event.hash, event.chainId),
+        initiatedBy: event.from as Address,
+        spentInGas: {
+          amount: event.receipt?.gasUsed || 0n,
+          amountInUnits: formatCurrencyAmount(event.receipt?.gasUsed || 0n, nativeCurrencyToken),
+        },
+        nativePrice: parseFloat(formatUnits(event.receipt?.effectiveGasPrice || 0n, nativeCurrencyToken.decimals)),
```
FiboApe, line 644: wait this is not the nativePrice. effectiveGasPrice is to calculate how much ETH was spent on the transaction. so actually spentInGas amount is effectiveGasPrice * gasUsed

FiboApe: We should try to get the native price from the balance state

apps/root/src/common/utils/transaction-history/parsing.tsx
```diff
@@ -592,4 +596,336 @@ const parseMultipleTransactionApiEventsToTransactionEvents = (
...
+          type: TransactionEventTypes.ERC20_APPROVAL,
+          data: {
+            token: { ...approvedToken, icon: <TokenIcon size={5} token={approvedToken} /> },
+            amount: {
+              amount,
+              amountInUnits,
+            },
+            owner: event.from as Address,
+            spender: event.typeData.addressFor as Address,
+            status: TransactionStatus.PENDING,
```
FiboApe, line 677: If they have a receipt they would not be pending right?

apps/root/src/common/utils/transactions.ts
```diff
@@ -67,3 +71,117 @@ export const getImpactedTokenForOwnWallet = (
...
+    },
+    remainingSwaps: BigInt(newPositionTypeData.frequencyValue),
+    totalSwaps: BigInt(newPositionTypeData.frequencyValue),
+    totalExecutedSwaps: 0n,
+    id,
+    startedAt: newPositionTypeData.startedAt,
+    pendingTransaction,
+    status: 'ACTIVE',
+    version: LATEST_VERSION,
+    isStale: newPositionTypeData.isStale,
```
FiboApe, line 161: this would always be false

apps/root/src/frame/components/promises-initializer/index.tsx
```diff
@@ -84,6 +84,9 @@ const PromisesInitializer = () => {
       timeoutPromise(contactListService.initializeAliasesAndContacts(), TimeoutPromises.COMMON, {
         description: ApiErrorKeys.LABELS_CONTACT_LIST,
       }).catch(handleError);
+      timeoutPromise(transactionService.fetchDcaIndexingBlocks(), TimeoutPromises.COMMON, {
+        description: ApiErrorKeys.HISTORY,
```
FiboApe, line 88: This should be another key

apps/root/src/services/positionService.ts
```diff
@@ -1669,17 +1579,22 @@ export default class PositionService extends EventsManager<PositionServiceData>
...
+          const newPosition = getNewPositionFromTxTypeData({
+            chainId: transaction.chainId,
             id: `${transaction.chainId}-${newId}-v${newPositionTypeData.version}`,
             positionId: BigInt(newId),
+            newPositionTypeData,
+            user: transaction.from as Address,
+          });
+
+          currentPositions[`${transaction.chainId}-${newId}-v${newPositionTypeData.version}`] = {
+            ...currentPositions[`pending-transaction-${transaction.hash}-v${newPositionTypeData.version}`],
```
FiboApe, line 1596: This might not exist if you enter the app after the transaction is confirmed but not indexed

apps/root/src/state/transactions/hooks.ts
```diff
@@ -343,3 +353,43 @@ export function useCampaignHasConfirmedTransaction(campaignId: string): boolean
...
+        } else {
+          return (
+            transaction.receipt.blockNumber >
+            BigInt(accountBlockNumbers[transaction.from as Address][chainId].processedUpTo)
+          );
+        }
+      });
+      return [...acc, ...filteredTransactions];
+    }, []);
+  }, [accountBlockNumbers, state]);
```
FiboApe, line 392: wallets should also be here

FiboApe: And the dcaIndexinBlocks

apps/root/src/state/transactions/hooks.ts
```diff
@@ -343,3 +353,43 @@ export function useCampaignHasConfirmedTransaction(campaignId: string): boolean
...
+            transaction.receipt.blockNumber > BigInt(dcaIndexingBlocks[chainId].processedUpTo)
+          );
+        } else {
+          return (
+            transaction.receipt.blockNumber >
+            BigInt(accountBlockNumbers[transaction.from as Address][chainId].processedUpTo)
+          );
+        }
+      });
+      return [...acc, ...filteredTransactions];
```
FiboApe, line 390: Lets go for modifying acc, so we don't shallow copy everytime

apps/root/src/state/transactions/transactionUpdater.tsx
```diff
@@ -122,19 +133,86 @@ export default function Updater(): null {
...
 
-    nonPendingTransactions.forEach((hash) => addTransactionToService(transactions[hash].receipt!, transactions[hash]));
-    dispatch(setInitialized());
+    nonPendingTransactions.forEach((tx) => addTransactionToService(tx.receipt!, tx));
 
+    dispatch(setInitialized());
     dispatch(setTransactionsChecking(pendingTransactions.map(({ hash, chainId }) => ({ hash, chainId }))));
   }, []);
 
+  useEffect(() => {
```
FiboApe, line 192: I'm not sure this should go in here, since it doesnt have anything to do with "handlingTransactions" it could be a redux thunk action that just dispatched the state and we call it right after the fetch of the positions

apps/root/src/hooks/useTransactionsHistory.tsx
```diff
@@ -73,7 +81,16 @@ function useTransactionsHistory(): {
...
-  }, [historyEvents, indexing, isLoadingTokenLists, isLoading, tokenList, nonIndexedTransactions, storedWallets]);
+  }, [
+    historyEvents,
+    indexing,
+    isLoadingTokenLists,
+    isLoading,
+    tokenList,
+    nonIndexedTransactions,
+    storedWallets,
+    nativePrices,
```
FiboApe, line 92: This doesnt impact performance right? You can easily check it with the activity and portfolio tables checking if they  keep rendering correctly or if there is a log between scrolling and the rows rendering

FiboApe (approved):
Looks good now!

PR #680, opened 2024-04-09
Merges main to privy integration

FiboApe:


apps/root/src/services/aggregatorService.ts
```diff
@@ -177,7 +177,7 @@ export default class AggregatorService {
       hasEnoughForSwap = sortedOptions.some((option) => balance >= option.sellAmount.amount);
     }
 
-    if (usePermit2 && from.address === protocolToken.address && takerAddress && hasEnoughForSwap) {
+    if (usePermit2 && from.address === protocolToken.address && takerAddress && hasEnoughForSwap && isOnNetwork) {
```
th0rOdinson, line 180: We agreed to implement validations no matter isOnNetwork value. Maybe we can merge this (as it's stable), and create a new task for it? @FiboApe

apps/root/src/constants/yield.ts
```diff
@@ -620,6 +621,50 @@ export const ALLOWED_YIELDS: Record<
...
+      id: '9a5f6bad-29d1-4fc2-b913-3adb5d394349', // Moonwell USDC.wh
+      tokenAddress: '0x487ddAdabFb26924878CA07d51F0Fe98114Fc511', // Moonwell USDC.wh
+      poolId: '9a5f6bad-29d1-4fc2-b913-3adb5d394349', // Moonwell USDC.wh
+      name: YieldName.moonwell,
+      token: emptyTokenWithAddress('MOONWELL'),
+    },
+    {
+      id: '48b2b90b-ab17-43a5-b7b0-2f91ec132d4d', // Moonwell xcUSDT
+      tokenAddress: '0xdf2ad201933c61Ec77b98d3800940342D8242906', // Moonwell xcUSDT
+      poolId: '48b2b90b-ab17-43a5-b7b0-2f91ec132d4d', // Moonwell xcUSDT
```
th0rOdinson, line 663: Just curious, there are two xcUSDT pools?

FiboApe: Seems like it, or the comment is wrong

th0rOdinson (approved):

PR #681, opened 2024-04-10
fix(dca): navigation and token list fixes

FiboApe:


packages/ui-library/src/components/token-picker/index.tsx
```diff
@@ -285,6 +286,31 @@ const TokenSearch = ({ search, onChange }: TokenSearchProps) => {
 
 const skeletonRows = Array.from(Array(10).keys()) as unknown as TokenWithBalance[];
 
+const removeIcon = (token: Token | TokenWithIcon): Token => {
+  // This breaks one of immers pitfalls and we store that into a redux state it slows everything down
```
th0rOdinson, line 290: Oh good catch!

th0rOdinson (approved):

PR #695, opened 2024-04-12
fix(history): decimal parsing error when parsing api events

th0rOdinson:
### Current behaviour
Screenshot 2024-04-12 at 11 35 19 AM

### New behaviour
Screenshot 2024-04-12 at 11 56 30 AM


FiboApe (approved):
ooops

PR #702, opened 2024-04-15
fix(*): refresh user data when adding new wallet

th0rOdinson:


apps/root/src/common/components/wallet-selector/index.tsx
```diff
@@ -105,6 +117,33 @@ const WalletSelector = ({ options, size = 'small' }: WalletSelectorProps) => {
     dispatch(cleanBalances());
   };
 
+  React.useEffect(() => {
+    const reFetchWalletsData = async () => {
```
FiboApe, line 121: wouldn't this also trigger the first time the user logs in?

th0rOdinson: It should not due to this condition:
`(prevWallets && prevWallets.length > 0 && wallets.length > prevWallets.length)`
We only update if there is already one wallet, and in this render we have more wallets than before

FiboApe (approved):

PR #710, opened 2024-04-16
fix(activity): show indexing wallet status

th0rOdinson:


apps/root/src/common/utils/transaction-history/index.ts
```diff
@@ -186,3 +186,34 @@ export const getTransactionPriceColor = (tx: TransactionEvent) => {
...
+  switch (tx.type) {
+    case TransactionEventTypes.ERC20_APPROVAL:
+      const { owner, spender } = tx.data;
+      wallets = [owner, spender];
+      break;
+    case TransactionEventTypes.DCA_WITHDRAW:
+    case TransactionEventTypes.DCA_TERMINATED:
+    case TransactionEventTypes.DCA_MODIFIED:
+    case TransactionEventTypes.DCA_PERMISSIONS_MODIFIED:
+    case TransactionEventTypes.DCA_CREATED:
```
FiboApe, line 201: DCA_CREATED could have a different user to the one that initiated the tx

FiboApe: Not through our UI, but it is available through SC so an event could come with those two values different

apps/root/src/hooks/useIsSomeWalletIndexed.ts
```diff
@@ -0,0 +1,27 @@
...
+
+const INDEXER_ACCEPTANCE = 0.95; // 95%
+
+export default function useIsSomeWalletIndexed(wallet?: Address) {
+  const { history } = useStoredTransactionHistory();
+
+  const isSomeWalletIndexed = React.useMemo(() => {
+    const returnValue = Object.entries(history?.indexing || {}).some(([address, chainsData]) => {
+      if (!wallet || wallet === address) {
+        const isSomeChainIndexed = Object.values(chainsData).some((chainData) => {
```
FiboApe, line 13: you can just return this

apps/root/src/hooks/useIsSomeWalletIndexed.ts
```diff
@@ -0,0 +1,27 @@
...
+import { Address } from 'viem';
+
+const INDEXER_ACCEPTANCE = 0.95; // 95%
+
+export default function useIsSomeWalletIndexed(wallet?: Address) {
+  const { history } = useStoredTransactionHistory();
+
+  const isSomeWalletIndexed = React.useMemo(() => {
+    const returnValue = Object.entries(history?.indexing || {}).some(([address, chainsData]) => {
+      if (!wallet || wallet === address) {
```
FiboApe, line 12: should return false if this is false right?

apps/root/src/hooks/useIsSomeWalletIndexed.ts
```diff
@@ -0,0 +1,27 @@
...
+import useStoredTransactionHistory from './useStoredTransactionHistory';
+import { Address } from 'viem';
+
+const INDEXER_ACCEPTANCE = 0.95; // 95%
+
+export default function useIsSomeWalletIndexed(wallet?: Address) {
+  const { history } = useStoredTransactionHistory();
+
+  const isSomeWalletIndexed = React.useMemo(() => {
+    const returnValue = Object.entries(history?.indexing || {}).some(([address, chainsData]) => {
```
FiboApe, line 11: Also here, you could return this directlty

apps/root/src/common/utils/transaction-history/index.ts
th0rOdinson: Pretty cool!

FiboApe (approved):

PR #713, opened 2024-04-16
fix(transfer): improve transfer to take into account users network an…

FiboApe:
…d wallet

### DIFF
--- a/apps/root/src/pages/transfer/components/transfer-button/index.tsx
+++ b/apps/root/src/pages/transfer/components/transfer-button/index.tsx
@@ -1,0 +1,70 @@
+import { NETWORKS } from '@constants';
+import useActiveWallet from '@hooks/useActiveWallet';
+import useCurrentNetwork from '@hooks/useCurrentNetwork';
+import useTrackEvent from '@hooks/useTrackEvent';
+import useWalletService from '@hooks/useWalletService';
+import { setNetwork } from '@state/config/actions';
+import { useAppDispatch } from '@state/hooks';
+import { useTransferState } from '@state/transfer/hooks';
+import { NetworkStruct } from 'common-types';
+import { find } from 'lodash';
+import React from 'react';
+import { FormattedMessage } from 'react-intl';
+import { Button } from 'ui-library';
+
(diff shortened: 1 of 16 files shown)


th0rOdinson (approved):

apps/root/src/pages/transfer/components/transfer-button/index.tsx
```diff
@@ -0,0 +1,70 @@
...
+    <Button fullWidth onClick={onTransferClick} disabled={disableTransfer} variant="contained">
+      {disableTransfer ? (
+        <FormattedMessage description="enterAmount" defaultMessage="Enter an amount" />
+      ) : (
+        <FormattedMessage description="transfer transferButton" defaultMessage="Transfer" />
+      )}
+    </Button>
+  );
+
+  const IncorrectNetworkButton = (
```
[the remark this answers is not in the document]
FiboApe: Why would we? The convention is that for React components that are already "rendered" we use UpperCase on their names

apps/root/src/state/create-position/reducer.ts
```diff
@@ -61,11 +64,13 @@ export default createReducer(initialState, (builder) => {
...
-      state.fromYield = payload;
+    .addCase(setFromYield, (state, { payload: { option, manualUpdate } }) => {
+      state.fromYield = option;
+      state.userHasChangedYieldOption = manualUpdate && true;
     })
-    .addCase(setToYield, (state, { payload }) => {
-      state.toYield = payload;
+    .addCase(setToYield, (state, { payload: { option, manualUpdate } }) => {
+      state.toYield = option;
+      state.userHasChangedYieldOption = manualUpdate && true;
```
th0rOdinson, line 73: Not sure why the && true is needed

th0rOdinson (approved):

apps/root/src/state/create-position/reducer.ts
FiboApe: I want to update it to true only if manualUpdate is true, but as I'm writing this I realize it is not necessary hehe

PR #717, opened 2024-04-18
Bly 2181

th0rOdinson:


apps/root/src/hooks/useTransactionsHistory.tsx
```diff
@@ -38,12 +38,8 @@ function useTransactionsHistory(): {
   const [parsedEvents, setParsedEvents] = React.useState<TransactionEvent[]>([]);
   const isLoading = isHookLoading || isLoadingService;
   const tokenList = useTokenList({});
-  const nonIndexedTransactions = useTransactionsAfterBlockNumber(indexing);
-
-  const chainsWithNativePrice = React.useMemo(
-    () => nonIndexedTransactions.map((tx) => tx.chainId),
-    [nonIndexedTransactions]
-  );
+  const pendingTransactions = usePendingTransactions();
```
FiboApe, line 41: Wait what about tx's that are confirmed but not indexed?

apps/root/src/state/transactions/actions.ts
```diff
@@ -47,6 +50,94 @@ export const cleanTransactions = createAction<{
   indexing: TransactionsHistory['indexing'];
 }>('transactions/cleanTransactions');
 
+export const fetchTransactionsHistory = createAppAsyncThunk<void, number | undefined>(
+  'transactions/fetchTransactionsHistory',
+  async (beforeTimestamp, { extra: { web3Service }, dispatch }) => {
+    const accountService = web3Service.getAccountService();
```
FiboApe, line 56: why did you need to remove this from the transaction history?

apps/root/src/hooks/useTransactionsHistory.tsx
```diff
@@ -54,22 +51,35 @@ function useTransactionsHistory(): {
         storedWallets.map(({ address }) => address)
       );
 
+      const nonIndexedTransactions = localTransactions.filter(
+        (tx) =>
+          !resolvedEvents.find(
+            (serviceEvent) =>
+              serviceEvent.tx.chainId === tx.chainId && serviceEvent.tx.txHash.toLowerCase() === tx.hash.toLowerCase()
```
FiboApe, line 58: shouldn't this be parsed with resolvedEvents? since its the final list we are going to show

apps/root/src/hooks/useTransactionsHistory.tsx
```diff
@@ -54,22 +51,35 @@ function useTransactionsHistory(): {
...
+          )
+      );
+      console.log('nonIndexedTransactions', nonIndexedTransactions);
```
FiboApe, line 61: 👁️

apps/root/src/hooks/useTransactionsHistory.tsx
```diff
@@ -54,22 +51,35 @@ function useTransactionsHistory(): {
...
       // This set parsed event is actually killing perf, making this hook re-render a thousand times infinitely
       // Haven't figure out just why this happens, as of right now this is a PATCH for it but not a definitive solution
       if (!isEqual(parsedEvents, resolvedEvents)) {
         setParsedEvents(resolvedEvents);
       }
 
-      if (indexing) dispatch(cleanTransactions({ indexing }));
+      const indexedTransactionsSummary = historyEvents.map(({ tx: { chainId, txHash } }) => ({ chainId, txHash }));
+
+      if (!!localTransactions.length) dispatch(cleanTransactions({ indexedTransactions: indexedTransactionsSummary }));
```
FiboApe, line 82: this can be a really big dispatch, what about just sending the last blockNumbers of the last transactions of each chain?

th0rOdinson: This would be a large list eventually, but each indexer unit has it's own last block number. We could do this calculation in the hook and send only the transactions to clean. We are kind of doing that already with nonIndexedTransactions

apps/root/src/hooks/useTransactionsHistory.tsx
th0rOdinson: Absolutely!

apps/root/src/hooks/useTransactionsHistory.tsx
FiboApe: Yup that works for me, and we only need to send the chainId and txHash :)

FiboApe (approved):
Looks good, heavily test this

apps/root/src/state/transactions/transactionUpdater.tsx
```diff
@@ -123,43 +126,111 @@ export default function Updater(): null {
...
-        eventToSearch: 'Deposited',
-      });
+    const getTokenWithPrice = async (token: Token) => {
+      const tokenBasePrice = await priceService.getUsdHistoricPrice([token], undefined, token.chainId);
+      const tokenPrice = parseUsdPrice(token, 10n ** BigInt(token.decimals), tokenBasePrice[token.address]);
+      return { ...token, price: tokenPrice };
+    };
 
-      if ('positionId' in parsedLog.args) {
+    switch (tx.type) {
```
FiboApe, line 139: This should all go inside a try catch, so as to not block that we confired a transaction if fetching the price fails

apps/root/src/state/transactions/transactionUpdater.tsx
```diff
@@ -195,7 +266,10 @@ export default function Updater(): null {
         .then(async (receipt) => {
           const tx = transactions[hash];
           if (receipt && !tx.receipt && receipt.status === 'success') {
-            const extendedTypeData = await parseTxExtendedTypeData(tx);
+            const extendedTypeData = await parseTxExtendedTypeData({
+              ...tx,
+              receipt: { ...receipt, chainId: tx.chainId },
```
FiboApe, line 271: do we need the chainId on the receipt?

FiboApe (approved):

apps/root/src/state/transactions/transactionUpdater.tsx
```diff
@@ -130,46 +126,118 @@ export default function Updater(): null {
...
+                to: withdrawPositionTokenWithPrice,
+              },
+            };
+          }
+          break;
+        case TransactionTypes.swap:
+          const swapTokenWithPrice = await getTokenWithPrice(tx.typeData.to);
+
+          extendedTypeData = {
+            to: swapTokenWithPrice,
```
FiboApe, line 223: wait aren't we missing the from here?

FiboApe (approved):

PR #729, opened 2024-04-23
Bly 2180

th0rOdinson:
### Demo
https://github.com/Mean-Finance/dca-fe/assets/144699868/ebed099b-2cb1-4f1d-931d-b0041c72eb82

### DIFF
--- a/packages/ui-library/src/components/transaction-confirmation/index.tsx
+++ b/packages/ui-library/src/components/transaction-confirmation/index.tsx
@@ -31,23 +31,10 @@ import {
 } from '../../emojis';
 
 const StyledOverlay = styled.div`
-  ${({
-    theme: {
-      palette: { mode },
-      spacing,
-    },
-  }) => `
-    position: absolute;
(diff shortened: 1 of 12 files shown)


packages/ui-library/src/components/transaction-confirmation/index.tsx
```diff
@@ -31,23 +31,10 @@ import {
...
-    bottom: 0;
-    left: 0;
-    right: 0;
-    z-index: 99;
-    padding: ${spacing(6)};
+  ${({ theme: { spacing } }) => `
     display: flex;
     flex-direction: column;
     gap: ${spacing(6)};
-    background-color: ${colors[mode].background.quarteryNoAlpha};
```
FiboApe, line 50: sure we should remove this one?

th0rOdinson: Yep. Components that used to be 'behind' are now not rendered

PR #734, opened 2024-04-24
Bly 2207

th0rOdinson:


apps/root/src/state/create-position/reducer.ts
```diff
@@ -78,6 +79,15 @@ export default createReducer(initialState, (builder) => {
     .addCase(setModeType, (state, { payload }) => {
       state.modeType = payload;
     })
+    .addCase(setDCADefaultChainId, (state, { payload }) => {
```
FiboApe, line 82: What's the difference from this to setDCAChainId?

apps/root/src/state/create-position/reducer.ts
```diff
@@ -78,6 +79,15 @@ export default createReducer(initialState, (builder) => {
...
+      state.fromValue = '';
+      state.from = null;
+      state.to = null;
+      state.fromYield = null;
+      state.toYield = null;
+      state.userHasChangedYieldOption = false;
+    })
     .addCase(setDCAChainId, (state, { payload }) => {
       state.chainId = payload;
       state.fromValue = '';
```
FiboApe, line 93: we need to modify the frequencyValues here and on resetDcaForm

apps/root/src/state/create-position/reducer.ts
th0rOdinson: setDCAChainID also resets the whole form. And I needed frequencyValue to be persistent. Now I'm thinking that this can be one action as it was, and persist frequency value an type even if network changes

FiboApe (approved):

PR #736, opened 2024-04-25
fix(transactionModal): handle user rejected transaction

th0rOdinson:
https://github.com/Mean-Finance/dca-fe/assets/144699868/71e30f69-e51f-441f-922f-84de092ffde7

### DIFF
--- a/apps/root/src/common/utils/errors.tsx
+++ b/apps/root/src/common/utils/errors.tsx
@@ -2,6 +2,7 @@ import React from 'react';
 import { FormattedMessage } from 'react-intl';
 import { Link } from 'ui-library';
 import styled from 'styled-components';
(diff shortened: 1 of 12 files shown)


apps/root/src/common/utils/errors.tsx
```diff
@@ -60,3 +64,25 @@ export const shouldTrackError = (error?: { code?: string; message?: string; reas
 
   return true;
 };
+
+const getTransactionExecutionErrorCode = (error: TransactionExecutionErrorType) => {
```
FiboApe, line 68: Carefull that not all errors are going to be caused by viem, we could have an error on our side, we should not treat every error as a viem one, but It might be one

th0rOdinson: I undestand. I've made some changes in order to make this more flexible

FiboApe (approved):

PR #739, opened 2024-04-25
fix(*): reconnect wallet

th0rOdinson:


apps/root/src/hooks/useOpenConnectModal.ts
```diff
@@ -11,71 +11,74 @@ const useOpenConnectModal = () => {
...
-          loginWalletTitle = defineMessage({
-            defaultMessage: 'Select your Wallet to link it.',
+          let loginWalletTitle = defineMessage({
+            defaultMessage: 'Select your Wallet to Log In.',
             description: 'RainbowWalletTitle',
           });
-        }
 
-        const titleContainer = document.querySelector('[data-testid="rk-connect-header-label"]');
+          if (user?.status === UserStatus.loggedIn && !isReconnecting) {
```
FiboApe, line 56: Are you formatting documents manually?

apps/root/src/common/components/wallet-selector/index.tsx
```diff
@@ -223,6 +223,20 @@ const WalletSelector = ({ options, size = 'small' }: WalletSelectorProps) => {
     type: OptionsMenuOptionType.option,
   };
 
+  const reconnectWalletOption: OptionsMenuOption = {
```
FiboApe, line 226: I think this should be an option of the button and not the wallet selector, plus, does this actually fix the issue of the wallet appearing as '...'?

apps/root/src/hooks/useOpenConnectModal.ts
th0rOdinson: WTF

th0rOdinson: Oh this happened because I added a parameter to the function, so everything is idented

apps/root/src/pages/transfer/components/transfer-button/index.tsx
```diff
@@ -21,10 +24,19 @@ const TransferButton = ({ disableTransfer, onTransferClick }: TransferButtonProp
   const { network } = useTransferState();
   const actualCurrentNetwork = useCurrentNetwork();
   const activeWallet = useActiveWallet();
+  const wallets = useWallets();
   const walletService = useWalletService();
   const dispatch = useAppDispatch();
   const trackEvent = useTrackEvent();
   const isOnCorrectNetwork = actualCurrentNetwork.chainId === network;
+  const openConnectModal = useOpenConnectModal();
+  const { disconnect } = useDisconnect({
```
FiboApe, line 33: This seems to be something we do everytime we open the connect modal, so probably should go inside the useOpenConnectModal hook

apps/root/src/pages/transfer/components/transfer-button/index.tsx
```diff
@@ -56,10 +76,18 @@ const TransferButton = ({ disableTransfer, onTransferClick }: TransferButtonProp
     </Button>
   );
 
+  const ReconnectButton = (
+    <Button fullWidth variant="contained" onClick={onConnectWallet}>
+      <FormattedMessage description="reconnect wallet" defaultMessage="Reconnect wallet" />
```
FiboApe, line 81: What I meant was having the account that needs to reconnect here

apps/root/src/hooks/useOpenConnectModal.ts
```diff
@@ -4,78 +4,112 @@ import useUser from './useUser';
...
+            loginDisclaimer = defineMessage({
+              defaultMessage: 'Reconnect to be able to operate with your Balmy account!',
+              description: 'RainbowReconnectWalletDisclaimer',
+            });
+            loginWalletTitle = defineMessage({
+              defaultMessage: 'Select your Wallet to reconnect to.',
+              description: 'RainbowReconnectWalletTitle',
+            });
+
+            reconnectWalletOptions = wallets
```
FiboApe, line 82: lets remove this, it does not look good

apps/root/src/pages/home/components/portfolio/index.tsx
```diff
@@ -140,7 +126,7 @@ const PortfolioNotConnected = () => {
           />
         </Typography>
       </ContainerBox>
-      <Button variant="contained" size="large" onClick={onConnectWallet} fullWidth>
+      <Button variant="contained" size="large" onClick={() => openConnectModal()} fullWidth>
```
FiboApe, line 129: same, no need to send it as a scope function

apps/root/src/pages/transfer/components/transfer-button/index.tsx
```diff
@@ -56,10 +62,25 @@ const TransferButton = ({ disableTransfer, onTransferClick }: TransferButtonProp
     </Button>
   );
 
+  const ReconnectButton = (
+    <Button fullWidth variant="contained" onClick={() => openConnectModal()}>
+      <FormattedMessage description="reconnect wallet" defaultMessage="Reconnect wallet" />
+      {authWallet && (
+        <>
+          {' ('}
+          <Address address={authWallet.address} trimAddress />
```
FiboApe, line 71: Same as before, plus we should be able to send this into the formatted message as a value

apps/root/src/common/components/wallet-selector/index.tsx
```diff
@@ -217,7 +204,7 @@ const WalletSelector = ({ options, size = 'small' }: WalletSelectorProps) => {
       })
     ),
     Icon: AddEmptyWalletIcon,
-    onClick: onConnectWallet,
+    onClick: () => openConnectModal(),
```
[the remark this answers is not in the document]
th0rOdinson: openConnectModal accepts a boolean parameter that is not compatible with native event parameter from HTMLButton onClick

apps/root/src/hooks/useOpenConnectModal.ts
```diff
@@ -4,78 +4,105 @@ import useUser from './useUser';
...
+  const { openConnectModal: openRainbowConnectModal } = useConnectModal();
   const user = useUser();
   const intl = useIntl();
   const mode = useThemeMode();
+  const wallets = useWallets();
+  const activeWallet = useActiveWallet();
 
-  return React.useCallback(() => {
-    if (!openConnectModal) return;
+  const isReconnecting = !!wallets.length && !activeWallet?.address;
```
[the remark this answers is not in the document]
th0rOdinson: We would be repeating this same conditional in every case. There is one case that depends on the consumer that is the position-list case, where the user is connected correctly, but we want to show this texts anyway

FiboApe (approved):

PR #743, opened 2024-04-26
Add much needed tracking

FiboApe:


apps/root/src/common/components/contact-modal/index.tsx
```diff
@@ -267,16 +269,29 @@ const ContactModal = ({
...
 
+  const onShowAddContact = () => {
+    setActiveModal(ContactListActiveModal.ADD_CONTACT);
+    trackEvent('Contact moda; - Open add contact', {
+      activeModal,
+    });
+  };
+
+  const onCloseModal = () => {
+    setActiveModal(ContactListActiveModal.ADD_CONTACT);
```
th0rOdinson, line 280: ContactListActiveModal.NONE !

apps/root/src/pages/dca/create-position/components/swap/index.tsx
```diff
@@ -364,7 +364,19 @@ const Swap = ({ currentNetwork, yieldOptions, isLoadingYieldOptions, handleChang
         shouldEnableYield ? toYield?.tokenAddress : undefined,
         signature
       );
-      trackEvent('DCA - Create position submitted');
+      try {
+        trackEvent('DCA - Create position submitted', {
```
th0rOdinson, line 368: You have added a try catch in the upper level, I don't think its necessary here

apps/root/src/pages/transfer/components/confirm-transfer-modal/index.tsx
```diff
@@ -116,13 +120,24 @@ const ConfirmTransferModal = ({
       const isProtocolToken = token.address === PROTOCOL_TOKEN_ADDRESS;
       const parsedToken: Token = { ...token, type: isProtocolToken ? TokenType.NATIVE : TokenType.ERC20_TOKEN };
 
+      trackEvent('Transfer - Transfer submitting', {
+        token: parsedToken,
+        isRecipientContact: contacts.find(({ address }) => address.toLowerCase() === recipient.toLowerCase()),
```
th0rOdinson, line 125: could calculate this once maybe?

apps/root/src/pages/dca/create-position/components/swap/index.tsx
FiboApe: Yeah but it parses the usd price at some point, and honestly, I was scared that it could explode through that, just being precacious

th0rOdinson (approved):

PR #751, opened 2024-04-29
feat(network-selector): improve selector height and order by net worth

FiboApe:


apps/root/src/pages/home/components/portfolio/index.tsx
```diff
@@ -302,6 +306,8 @@ const Portfolio = ({ selectedWalletOption }: PortfolioProps) => {
 
   const isLoading = isLoadingAllBalances || isLoggingUser;
 
+  const intlContext = React.useMemo(() => ({ intl }), [intl]);
```
th0rOdinson, line 309: Just curious why this needs memo

th0rOdinson (approved):

PR #761, opened 2024-04-29
Bly 2286

th0rOdinson:


apps/root/src/pages/dca/positions/components/positions-list/current-positions/components/position-card-button/index.tsx
```diff
@@ -117,11 +118,19 @@ const PositionCardButton = ({
...
+  const reconnectingWalletDisplay = getDisplayWallet(wallet);
+
   return (
     <StyledCallToActionContainer>
       {!walletIsConnected && (
         <StyledCardFooterButton onClick={openConnectModal} fullWidth size="large">
-          <FormattedMessage description="reconnect wallet" defaultMessage="Reconnect wallet" />
+          <FormattedMessage
+            description="reconnect wallet"
+            defaultMessage="Reconnect wallet{wallet}"
```
FiboApe, line 129: parenthesis and spaces should be here

packages/ui-library/src/theme/variants/button-variants.ts
```diff
@@ -34,7 +34,6 @@ export const buildButtonVariant = (mode: 'light' | 'dark'): Components => ({
         },
         '&:hover': {
           backgroundColor: colors[mode].accent.accent200,
-          border: `1.5px solid ${colors[mode].accent.accent400}`,
           color: colors[mode].accent.accent600,
```
FiboApe, line 37: what about on contained on the bottom, the borderWidth: '1.5px'?

apps/root/src/pages/dca/positions/components/positions-list/current-positions/components/position-card-button/index.tsx
th0rOdinson: This is just in case so we dont show Reconnect wallet (), because wallet could be undefined

PR #774, opened 2024-04-30
fix(transfer): validate recipient in edge cases

th0rOdinson:
With this change:

- Contact as recipient is being recognized
- Active wallet does not appear as contact
- Clicking a contact is validated as same as typing the address (handy when active wallet is changed, after the contact is selected)
- Url param is now being validated too

### DIFF
--- a/apps/root/src/hooks/useStoredContactList.ts
+++ b/apps/root/src/hooks/useStoredContactList.ts
@@ -6,9 +6,11 @@ import useStoredLabels from './useStoredLabels';
 import useWallets from './useWallets';
 import { map, uniq } from 'lodash';
 import { ContactList } from 'common-types';
+import useActiveWallet from './useActiveWallet';
 
 function useStoredContactList() {
   const contactListService = useContactListService();
+  const activeWallet = useActiveWallet();
   const labels = useStoredLabels();
   const wallets = useWallets();
 
@@ -19,7 +21,9 @@ function useStoredContactList() {
 
   const labeledContactList = React.useMemo<ContactList>(() => {
     const contactsWithWallets = uniq([...map(storedContactList, 'address'), ...map(wallets, 'address')]);
-    return contactsWithWallets.map((address) => ({ address, label: labels[address] }));
+    return contactsWithWallets
+      .filter((address) => address !== activeWallet?.address)
+      .map((address) => ({ address, label: labels[address] }));
   }, [storedContactList, labels, wallets]);
 
   return labeledContactList;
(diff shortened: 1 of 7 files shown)


apps/root/src/hooks/useStoredContactList.ts
```diff
@@ -19,7 +21,9 @@ function useStoredContactList() {
 
   const labeledContactList = React.useMemo<ContactList>(() => {
     const contactsWithWallets = uniq([...map(storedContactList, 'address'), ...map(wallets, 'address')]);
-    return contactsWithWallets.map((address) => ({ address, label: labels[address] }));
+    return contactsWithWallets
+      .filter((address) => address !== activeWallet?.address)
```
FiboApe, line 25: Mmm I think it would be better to filter this on the contacts modal directly so we can use storedContactList everywhere to check if the address is already a contact regardless of the active wallet

apps/root/src/pages/transfer/components/recipient-address/index.tsx
```diff
@@ -7,22 +7,51 @@ import { useTransferState } from '@state/transfer/hooks';
...
+    const foundContact = contactList.find((contact) => contact.address === storedRecipient);
+
+    if (!foundContact) {
+      return;
+    }
+
+    if (!foundContact.label?.label) {
+      return intl.formatMessage(
+        defineMessage({
+          defaultMessage: 'This address matches a contact of yours',
```
FiboApe, line 35: Wait if we dont have a label is it really a contact? Or does this refer to one of the addresses you have added but dont have a label?

0xged:
While checking out the changes, I realized that we are showing the name of the contact, the way it should be shown IF the user INPUTS the same address than a contact:
image

While, as per Figma, we should show it this way if they SELECT a contact:
image


wdyt?

apps/root/src/pages/transfer/components/recipient-address/index.tsx
th0rOdinson: Second one 😄 . It could be a contact/linkedWallet without label. It seemed strange to show 'this address matches with 0x...'

FiboApe: yeah lets leave the text change for @0xged to handle, I would go with something like `This address matches one of your added wallets`

th0rOdinson: There is still a case where you can add a contact without a label

FiboApe (approved):

PR #784, opened 2024-05-03
fix(customTokens): fix across app

th0rOdinson:


apps/root/src/state/token-lists/actions.ts
```diff
@@ -66,7 +70,7 @@ export const fetchTokenDetails = createAppAsyncThunk<Token, { tokenAddress: stri
     ]);
 
     const customToken = toToken({
-      address: tokenAddress,
+      address: tokenAddress.toLowerCase(),
```
FiboApe, line 73: maybe this should be on the toToken utility

apps/root/src/state/balances/actions.ts
```diff
@@ -129,6 +83,70 @@ export const updateTokens = createAppAsyncThunk<void, { tokens: Token[]; chainId
   }
 );
 
+export const fetchInitialBalances = createAppAsyncThunk<BalancesState['balances'], void>(
+  'balances/fetchInitialBalances',
+  async (_, { dispatch, getState, extra: { web3Service } }) => {
+    const accountService = web3Service.getAccountService();
+    const meanApiService = web3Service.getMeanApiService();
+    const sdkService = web3Service.getSdkService();
+    const chainIds = sdkService.getSupportedChains();
```
FiboApe, line 92: This is just grabbing the aggregator supported chains

apps/root/src/hooks/useHandleCustomTokenBalances.ts
```diff
@@ -0,0 +1,33 @@
...
+      } catch (e) {
+        console.error(e);
+      } finally {
+        setIsLoadingCustomToken(false);
+      }
+    },
+    []
+  );
+
+  return { handleCustomTokenBalances, isLoadingCustomToken };
```
FiboApe, line 29: We should memo this

apps/root/src/hooks/useHandleCustomTokenBalances.ts
```diff
@@ -0,0 +1,33 @@
...
+}
+useHandleCustomTokenBalances.whyDidYouRender = true;
```
FiboApe, line 31: 👁️

apps/root/src/hooks/useHandleCustomTokenBalances.ts
```diff
@@ -0,0 +1,33 @@
...
+import { Address } from 'viem';
+import { useCustomTokens } from '@state/token-lists/hooks';
+import { getTokenListId } from '@common/utils/parsing';
+
+function useHandleCustomTokenBalances() {
+  const dispatch = useAppDispatch();
+  const [isLoadingCustomToken, setIsLoadingCustomToken] = React.useState(false);
+  const customTokens = useCustomTokens();
+
+  const handleCustomTokenBalances = React.useCallback(
```
FiboApe, line 13: Lets brainstorm another number

apps/root/src/pages/aggregator/swap-container/index.tsx
```diff
@@ -69,32 +79,28 @@ const SwapContainer = () => {
...
-          networkToUse = find(mappedNetworks, { name: chainId.toLowerCase() });
-        }
-        dispatch(
-          setFrom(
-            getProtocolToken(Number(networkToUse?.chainId || actualCurrentNetwork.chainId || currentNetwork.chainId))
-          )
-        );
-      }
+    const handleCustomTokens = (tokenAddress: Address) => {
+      void handleCustomTokenBalances({ tokenAddress, chainId: networkToUse });
```
FiboApe, line 83: This network should actually be the chainId passed by param right?

apps/root/src/pages/aggregator/swap-container/index.tsx
```diff
@@ -69,32 +79,28 @@ const SwapContainer = () => {
...
+      handleCustomTokens(fromParam);
+    }
+    if (!isLoadingAllTokenLists && !toParamToken && toParam && isAddress(toParam)) {
+      handleCustomTokens(toParam);
+    }
+  }, [isLoadingAllTokenLists]);
+
+  React.useEffect(() => {
+    if ((!from && fromParamToken) || (fromParamToken && prevIsLoadingCustomToken && !isLoadingCustomToken)) {
+      dispatch(setFrom(fromParamToken));
```
FiboApe, line 95: We might have some unwanted re-renders here if the custom token search is modified since its going to modify this on the top and the isLoading params are going to change, thus entering here again an dispatching

apps/root/src/pages/aggregator/swap-container/index.tsx
```diff
@@ -69,32 +79,28 @@ const SwapContainer = () => {
...
     }
 
-    if (!to) {
-      if (toParamToken) {
-        dispatch(setTo(toParamToken));
-      } else if (toParamCustomToken && !to) {
-        dispatch(setTo(toParamCustomToken.token));
-      }
+    if ((!to && toParamToken) || (toParamToken && prevIsLoadingCustomToken && !isLoadingCustomToken)) {
+      dispatch(setTo(toParamToken));
```
FiboApe, line 101: Same

apps/root/src/common/utils/parsing.ts
```diff
@@ -400,12 +400,16 @@ export const parseTokensForPicker = ({
   balances: TokenBalances;
   customTokens?: TokenList;
 }) => {
-  const tokenKeys = Object.keys(tokenList);
-  const customTokenKeys = Object.keys(customTokens || {});
+  const mergedTokenLists = {
+    ...tokenList,
+    ...customTokens,
```
FiboApe, line 405: but custom tokens can be undefined D:

FiboApe: btw I think the order should be reversed, so we replace any custom token with the one that we have on our list that should be more curated

apps/root/src/common/components/token-selector/index.tsx
```diff
@@ -112,31 +129,38 @@ const TokenSelector = ({ handleChange, selectedToken }: TokenSelectorProps) => {
...
-        ['desc']
-      ),
-    [availableTokens, customToken]
-  );
+  const items: OptionWithKeyAndToken[] = React.useMemo(() => {
+    const parsedWithCustomTokens = availableTokens.map((tokenOption) => ({
+      ...tokenOption,
+      token: {
+        ...tokenOption.token,
+        isCustomToken:
```
FiboApe, line 137: Carefull with this, if we later add that custom token to our list, this is still gonna show as a custom token since it would still be maintained in their local storage, when it is already curated by us

apps/root/src/common/components/token-selector/index.tsx
```diff
@@ -112,31 +129,38 @@ const TokenSelector = ({ handleChange, selectedToken }: TokenSelectorProps) => {
...
+          !!customTokens[`${tokenOption.token.chainId}-${tokenOption.token.address.toLowerCase()}` as TokenListId],
+      },
+    }));
+
+    return orderBy(
+      parsedWithCustomTokens,
+      [({ balanceUsd, token }) => parseFloat(formatUnits(balanceUsd || 0n, token.decimals + 18))],
```
FiboApe, line 144: we can use parseUsdPrice here right?

apps/root/src/pages/aggregator/swap-container/components/token-picker/index.tsx
```diff
@@ -28,74 +24,52 @@ const AggregatorTokenPicker = ({ shouldShow, onChange, onClose, modalTitle }: Ag
...
-      (customToken.balance && {
-        amount: customToken.balance,
-        amountInUnits: formatCurrencyAmount({ amount: customToken.balance, token: customToken.token }),
-        amountInUSD:
-          (customToken.balanceUsd &&
-            parseFloat(formatUnits(customToken.balanceUsd, customToken.token.decimals + 18)).toFixed(2)) ||
-          undefined,
-      }) ||
-      undefined,
+  // TODO: Track custom tokens addition like it was tracked before
```
FiboApe, line 46: lets add this :)

apps/root/src/pages/aggregator/swap-container/index.tsx
th0rOdinson: Yes, but as the param can be the network name, i've calculated it on 'networkToUse' useMemo

apps/root/src/pages/aggregator/swap-container/index.tsx
th0rOdinson: I have re-defined the customToken hook so we can instance it once per token. This useEffect should not be triggered if the search changes, it's only meant to handle the url params. So the if blocks are only executed when a custom token from url param was loading in a previous render, but now it's not loading anymore.

Also, the isLoading for custom tokens in this component should only be triggered once, when the tokenList has loaded

apps/root/src/common/utils/parsing.ts
th0rOdinson: Absolutely! I didn't considered this case, thank you

apps/root/src/common/components/token-selector/index.tsx
th0rOdinson: I've included a solution for this, but I'm thinking if we should have a cleanCustomToken action, like we have for transactions

apps/root/src/common/components/token-selector/index.tsx
th0rOdinson: With no price available in the component, I've choosen `formatUsdAmount` 😃

apps/root/src/pages/aggregator/swap-container/components/token-picker/index.tsx
th0rOdinson: I have 😄 but forgot to remove this comment (done in line 55 of this file)

apps/root/src/hooks/useAddCustomTokenToList.ts
```diff
@@ -0,0 +1,51 @@
...
+  const dispatch = useAppDispatch();
+  const [customTokenAddress, setCustomTokenAddress] = React.useState<Address | undefined>();
+  const [isLoadingCustomToken, setIsLoadingCustomToken] = React.useState(false);
+  const customTokens = useCustomTokens();
+  const selectedNetwork = useSelectedNetwork();
+  const trackEvent = useTrackEvent();
+
+  React.useEffect(() => {
+    const handleCustomTokenBalances = async (tokenAddress: Address) => {
+      const network = chainId || selectedNetwork.chainId;
```
FiboApe, line 20: selectedNetwork is not on the dependency list

FiboApe: Also customTokens is not on the dependencyList, and when both are on the dependency list we'll get a lot more fetching of the custom token due to re-renders caused by not the tokenAddress, so we probably need to check for isLoading or that the addresses are different etc

apps/root/src/state/token-lists/actions.ts
```diff
@@ -38,11 +38,16 @@ export const startFetchingTokenLists = createAppAsyncThunk(
...
+    const { byUrl: tokensLists, customTokens } = getState().tokenLists;
 
     const tokenList = parseTokenList({
-      tokensLists,
+      tokensLists: {
+        ...tokensLists,
+        'custom-tokens': customTokens,
+      },
+      chainId,
+      curateList: true,
```
FiboApe, line 50: I'm not sure we want to curate this here, since we are asking for a specific token we should probably return it either if it was curated or not

apps/root/src/hooks/useAddCustomTokenToList.ts
th0rOdinson: I'm not finding a case where we want to trigger this based on an update from network/customTokens. For our current cases, we are controlling this hook by just triggering it with a new tokenAddress

apps/root/src/common/components/token-selector/index.tsx
```diff
@@ -167,7 +167,7 @@ const TokenSelector = ({ handleChange, selectedToken }: TokenSelectorProps) => {
       isAddress(searchTerm) &&
       !tokens[`${selectedNetwork.chainId}-${searchTerm.toLowerCase()}` as TokenListId]
     ) {
-      setCustomTokenAddress(searchTerm);
+      void addCustomTokenToList(searchTerm);
```
FiboApe, line 170: missing chainId

packages/ui-library/src/components/token-picker/index.tsx
```diff
@@ -400,11 +400,15 @@ const TokenPicker = ({
...
-  useEffect(() => {
-    if (search && filteredTokens.length === 0 && validAddressRegex.test(search) && onFetchCustomToken) {
-      onFetchCustomToken(search.toLowerCase() as Address);
-    }
-  }, [search, filteredTokens, onFetchCustomToken]);
+  const onSearchChange = useCallback(
+    (newSearch: string) => {
+      setSearch(newSearch);
+      if (newSearch && validAddressRegex.test(newSearch) && onFetchCustomToken) {
+        onFetchCustomToken(newSearch.toLowerCase() as Address);
```
FiboApe, line 407: Missing chain id :)

th0rOdinson: I set chainId as optional so its can be calculated directly in the hook by default (using useSelectedNetwork), unless it needs to be specified (as it is in the case for the url param)

FiboApe: yeah but this token picker is opened in a specific chainId so we should also forward that so it doesnt find any inconsistencies

apps/root/src/state/balances/actions.ts
```diff
@@ -99,7 +100,21 @@ export const fetchInitialBalances = createAppAsyncThunk<BalancesState['balances'
...
+    // Merging api balances with customToken balances
+    const mergedBalances = cloneDeep(accountBalancesResponse.balances);
+    const customTokens = getState().tokenLists.customTokens.tokens;
+
+    for (const walletAddress of wallets) {
+      const customTokensBalances = await sdkService.getMultipleBalances(customTokens, walletAddress);
+
+      for (const [chainId, balanceList] of Object.entries(customTokensBalances)) {
+        for (const [tokenAddress, balance] of Object.entries(balanceList)) {
+          mergedBalances[walletAddress][Number(chainId)][tokenAddress] = balance.toString();
```
FiboApe, line 112: shouldn't we check that '[walletAddress][chainId]' existsfirst?

th0rOdinson: True! Thanks

FiboApe (approved):

PR #786, opened 2024-05-06
feat(balances): start using curated token list for balances updater

th0rOdinson:


apps/root/src/hooks/useTokenListByChainId.ts
```diff
@@ -1,46 +1,24 @@
...
-      return acc;
-    }, {});
-  }, [tokensLists, savedAllTokenLists, filter]);
-
-  return tokenListByChainId;
+  return React.useMemo(
+    () =>
+      Object.entries(tokenList).reduce<TokenListByChainId>(
+        (acc, [tokenKey, token]) => ({
+          ...acc,
```
FiboApe, line 12: This could bring down performance, simply by the amount of tokens we may handle. Lets move this function to modify acc directly

FiboApe (approved):

PR #797, opened 2024-05-08
fix(dca): handle protocol token based on api new implementation

th0rOdinson:


FiboApe (approved):
This looks good but lets wait to merge it till tomorrow that I want to test it extensively

PR #815, opened 2024-05-16
feat(txReceipt): modify position receipt

th0rOdinson:
Screenshot 2024-05-16 at 4 31 31 PM

### DIFF
--- a/apps/root/src/common/utils/transaction-history/parsing.tsx
+++ b/apps/root/src/common/utils/transaction-history/parsing.tsx
@@ -910,8 +910,7 @@ export const transformNonIndexedEvents = ({
         baseEventData = buildBaseDcaPendingEventData(position);
 
         const totalBefore = position.rate.amount * BigInt(position.remainingSwaps);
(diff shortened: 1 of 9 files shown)


apps/root/src/common/utils/transaction-history/parsing.tsx
```diff
@@ -885,7 +910,8 @@ export const transformNonIndexedEvents = ({
         baseEventData = buildBaseDcaPendingEventData(position);
 
         const totalBefore = position.rate.amount * BigInt(position.remainingSwaps);
-        const totalNow = BigInt(event.typeData.newRate) * BigInt(event.typeData.newSwaps);
+        const newRate = parseUnits(event.typeData.newRate, event.typeData.decimals);
```
FiboApe, line 913: were we actually receiving the value as "0.1" ?

th0rOdinson: Yep. After thinking about this, I finally parsed this in the component 😄

PR #829, opened 2024-05-24
feat(dashboard): allow hidding and showing balances

FiboApe:


apps/root/src/pages/home/components/portfolio/index.tsx
```diff
@@ -184,8 +192,14 @@ const PortfolioBodyItem: ItemContent<BalanceItem, Context> = (
               <Skeleton variant="text" animation="wave" sx={{ minWidth: '5ch' }} />
             ) : (
               <ContainerBox alignItems="center" gap={3}>
-                <CircularProgressWithBrackground thickness={8} size={SPACING(6)} value={relativeBalance} />
-                <StyledBodySmallLabelTypography>{relativeBalance.toFixed(0)}%</StyledBodySmallLabelTypography>
+                <CircularProgressWithBrackground
+                  thickness={8}
+                  size={SPACING(6)}
+                  value={showBalances ? relativeBalance : 0}
```
th0rOdinson, line 198: This got me thinking... Should this % be hidden? I don't think it's wrong to hide it, but I also don't consider it sensitive information

apps/root/src/state/index.ts
```diff
@@ -142,6 +143,7 @@ const createStore = (web3Service: Web3Service) =>
           theme: 'light',
           selectedLocale: SupportedLanguages.english,
           showSmallBalances: true,
+          showBalances: true,
```
th0rOdinson, line 146: As falsy values are being removed from localStorage, I suggest we save `config` as an object rather than storing its values separately. Otherwise, we would be always having showBalances as true by default, and never saving user's preference

apps/root/src/pages/home/components/activity/index.tsx
```diff
@@ -155,15 +157,17 @@ const ActivityContent: ItemContent<TransactionEvent, Context> = (
         <StyledBodySmallLabelTypography noWrap={false}>{formattedDate}</StyledBodySmallLabelTypography>
       </StyledOperation>
       <StyledValue>
-        <StyledBodySmallRegularTypo2 noWrap={false}>{txTokenFlow}</StyledBodySmallRegularTypo2>
+        <StyledBodySmallRegularTypo2 noWrap={false}>
+          {showBalances ? txTokenFlow : <HiddenNumber size="small" />}
+        </StyledBodySmallRegularTypo2>
```
th0rOdinson, line 162: The UI changes a little when toggling showBalances, I think we can solve it by changing <HiddenNumber to be outside the <Typography /> and increasing its line height a little to be 18px

https://github.com/Balmy-protocol/dca-fe/assets/144699868/bee26fba-81f3-4c13-a265-3b141ab1d5de

apps/root/src/pages/home/components/portfolio/index.tsx
FiboApe: Nope, at least based on the designs we want to show it

apps/root/src/state/index.ts
FiboApe: Are you sure they are being removed? The case where you had that happen had a problem where you didn't add it to the preloaded states

apps/root/src/pages/home/components/activity/index.tsx
FiboApe: I mean it's going to change based on that now the bottom date has more space to expand, I don't think it is bad

apps/root/src/state/index.ts
```diff
@@ -158,16 +128,6 @@ const createStore = (web3Service: Web3Service) =>
           ],
           byUrl: getDefaultByUrl(),
           hasLoaded: false,
-          customTokens: {
-            name: 'custom-tokens',
-            logoURI: '',
-            timestamp: new Date().getTime(),
-            tokens: [],
```
th0rOdinson, line 163: Removing this `customTokens` object throws an error when localStorage is fully empty

th0rOdinson (approved):

PR #833, opened 2024-05-29
feat(earn): All Strategies Table

th0rOdinson:
https://github.com/Balmy-protocol/dca-fe/assets/144699868/c1ec847c-8a2f-40ea-8d36-5fd213577d9d


apps/root/src/pages/earn/components/all-strategies-table/index.tsx
```diff
@@ -0,0 +1,211 @@
...
+
+  React.useEffect(() => {
+    const fetchStrategies = async () => {
+      await earnService.fetchAllStrategies();
+    };
+    void fetchStrategies();
+  }, []);
+
+  const visibleRows = React.useMemo<Strategy[]>(
+    () => strategies.slice(page * ROWS_PER_PAGE, page * ROWS_PER_PAGE + ROWS_PER_PAGE),
```
FiboApe, line 185: have you checked with the BE team that the strategies are actually not going to be paginated on the BE?

apps/root/src/pages/earn/components/all-strategies-table/index.tsx
```diff
@@ -0,0 +1,211 @@
...
+          <TableHead>
+            <AllStrategiesTableHeader />
+          </TableHead>
+          <TableBody>
+            {isLoadingAllStrategies ? <AllStrategiesTableBodySkeleton /> : visibleRows.map(createRow)}
+          </TableBody>
+        </Table>
+        <TablePagination
+          count={strategies.length}
+          rowsPerPage={10}
```
FiboApe, line 202: You can use the constant here :D

apps/root/src/pages/earn/components/all-strategies-table/index.tsx
```diff
@@ -0,0 +1,211 @@
...
+      <ContainerBox>
+        <StyledYieldTypeBox>
+          <StyledBodySmallRegularTypo2>{strategy.farm.yieldType}</StyledBodySmallRegularTypo2>
+        </StyledYieldTypeBox>
+      </ContainerBox>
+    </TableCell>
+    <TableCell>
+      <StyledBodySmallSemiboldTypo2>${usdFormatter(strategy.farm.tvl)}</StyledBodySmallSemiboldTypo2>
+    </TableCell>
+    <TableCell>
```
FiboApe, line 166: what about security?

packages/common-types/src/earn.ts
```diff
@@ -0,0 +1,44 @@
...
+  LENDING,
+  STAKING,
+}
+
+export type Strategy = {
+  id: string;
+  asset: TokenWithIcon;
+  rewards: { token: TokenWithIcon; apy: number }[];
+  network: NetworkStruct;
+  farm: Omit<ApiFarm, 'yieldType'> & { yieldType: string };
```
FiboApe, line 42: We should probably add a new attribute for this, and not replace the one there

apps/root/src/pages/earn/components/all-strategies-table/index.tsx
th0rOdinson: Wasn't sure if implement it already, but I can include it now 😄

packages/ui-library/src/theme/variants/table-variants.ts
```diff
@@ -35,4 +36,22 @@ export const buildTableVariant = (mode: 'light' | 'dark'): Components => ({
       },
     },
   },
+  MuiTablePagination: {
+    defaultProps: {
+      component: 'div',
+      rowsPerPageOptions: [],
```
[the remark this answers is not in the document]
th0rOdinson: By default mui shows a dropdown allowing the user to set how many items want to see per page (array values). I have set that node to display: 'none', but also keeping this option as an empty array sets that node as empty too

apps/root/src/common/utils/earn/parsing.tsx
```diff
@@ -0,0 +1,64 @@
...
+
+    return {
+      id: strategy.id,
+      asset: {
+        ...assetToken,
+        icon: <TokenIcon size={7} token={assetToken} />,
+      },
+      rewards: strategy.rewards.map((reward) => ({
+        token: {
+          ...tokenList[getTokenListId({ tokenAddress: reward.token, chainId: strategy.chainId })],
```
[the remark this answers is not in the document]
th0rOdinson: Absolutely! BE is going to add more fields to this response, so I will improve this part now

packages/common-types/src/earn.ts
```diff
@@ -29,6 +29,14 @@ type ApiGuardianFee = {
   percentage: number;
 };
 
+export type ApiEarnToken = {
```
FiboApe, line 32: We do have this type somewhere right? or is it exported from the sdk directly?

FiboApe (approved):

packages/common-types/src/earn.ts
th0rOdinson: We have `DCAPositionToken`:
```
export type DCAPositionToken = {
    address: TokenAddress;
    symbol: string;
    name: string;
    decimals: number;
    price?: number;
    variant: TokenVariant;
};
I supposed we would need a new type for earn, so I created this new `ApiEarnToken`

FiboApe: Mmm, lets check if we can import it from the sdk directly, if not merge it as is

PR #834, opened 2024-05-29
fix(agg): Allow non-connected wallets to get swap options

th0rOdinson:


apps/root/src/hooks/useSwapOptions.ts
```diff
@@ -85,23 +85,23 @@ function useSwapOptions(
...
-              debouncedGasSpeed,
-              debouncedAccount as Address,
-              debouncedChainId,
-              debouncedDisabledDexes,
-              // eslint-disable-next-line @typescript-eslint/no-unsafe-argument
-              debouncedIsPermit2Enabled,
-              // eslint-disable-next-line @typescript-eslint/no-unsafe-argument
-              debouncedSourceTimeout
-            );
+            const promiseResult = await aggregatorService.getSwapOptions({
```
0xsambugs, line 88: Really like using the named params 👏

apps/root/src/hooks/useTotalGasSaved.ts
```diff
@@ -37,18 +37,13 @@ function useTotalGasSaved(position: Position | undefined | null): [bigint | unde
...
-            position.rate.amount,
-            undefined,
-            SORT_LEAST_GAS,
-            undefined,
-            undefined,
-            undefined,
-            undefined,
-            position.chainId
-          );
+          const options = await aggregatorService.getSwapOptions({
```
0xsambugs, line 40: Same!

apps/root/src/services/aggregatorService.ts
```diff
@@ -91,28 +91,45 @@ export default class AggregatorService {
...
+    usePermit2?: boolean;
+    sourceTimeout?: TimeoutKey;
+  }) {
     const balance = await this.walletService.getBalance({ account: takerAddress, token: from });
 
-    const isOnNetwork = !chainId || currentNetwork.chainId === chainId;
+    let isOnNetwork = false;
+    try {
+      const currentNetwork = await this.providerService.getNetwork(takerAddress);
+      isOnNetwork = !chainId || currentNetwork.chainId === chainId;
```
0xsambugs, line 128: It is worse for readability, but if `!chainId`, then you don't need to ask the provider service for the network, and you don't need to wait for it
```suggestion
      isOnNetwork = !chainId || (await this.providerService.getNetwork(takerAddress)).chainId === chainId;
```

0xsambugs (approved):
Looks good!

apps/root/src/services/aggregatorService.ts
FiboApe: Agree

th0rOdinson: Like it!

FiboApe (approved):

PR #835, opened 2024-05-30
feat(earn): Earn FAQ structure

th0rOdinson:


apps/root/src/pages/earn/components/faq/index.tsx
```diff
@@ -0,0 +1,183 @@
...
+            <StyledQuestion>
+              <FormattedMessage
+                description="faqWhatIsMeanFinance"
+                defaultMessage="What is Dollar Cost Averaging (DCA)?"
+              />
+            </StyledQuestion>
+          </AccordionSummary>
+          <AccordionDetails>
+            <StyledAnswer>
+              <FormattedMessage
```
FiboApe, line 51: Lets leave only one of this and with a diferent description in case we change it on the original one so it does not impact translations

FiboApe (approved):

PR #839, opened 2024-06-05
feat(earn): search for all strategies table

th0rOdinson:
https://github.com/Balmy-protocol/dca-fe/assets/144699868/faa04cc1-8f0b-4f90-96f8-3f96e497cef2

### DIFF
--- a/apps/root/src/pages/earn/components/all-strategies-table/components/toolbar/index.tsx
+++ b/apps/root/src/pages/earn/components/all-strategies-table/components/toolbar/index.tsx
@@ -3,16 +3,18 @@ import { useThemeMode } from '@state/config/hooks';
 import { FormattedMessage, defineMessage, useIntl } from 'react-intl';
 import { ContainerBox, InputAdornment, SearchIcon, TextField, Typography, colors } from 'ui-library';
 import TableFilters from '../filters';
+import { useAllStrategiesFilters } from '@state/all-strategies-filters/hooks';
 
 interface AllStrategiesTableToolbarProps {
   isLoading: boolean;
-  search: string;
-  setSearch: (newValue: string) => void;
+  setSearch: (search: string) => void;
 }
 
-const AllStrategiesTableToolbar = ({ isLoading, search, setSearch }: AllStrategiesTableToolbarProps) => {
+const AllStrategiesTableToolbar = ({ isLoading, setSearch }: AllStrategiesTableToolbarProps) => {
   const intl = useIntl();
   const themeMode = useThemeMode();
+  const { search } = useAllStrategiesFilters();
+
   return (
     <ContainerBox justifyContent="space-between" alignItems="end">
       <Typography variant="h4Bold" color={colors[themeMode].typography.typo1}>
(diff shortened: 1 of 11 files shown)


apps/root/src/pages/earn/components/all-strategies-table-toolbar/index.tsx
```diff
@@ -0,0 +1,54 @@
+import React from 'react';
+import { useThemeMode } from '@state/config/hooks';
+import { FormattedMessage, defineMessage, useIntl } from 'react-intl';
+import styled from 'styled-components';
+import { ContainerBox, InputAdornment, SearchIcon, TextField, Typography, colors } from 'ui-library';
+
+const StyledSearchBox = styled(TextField)`
```
FiboApe, line 7: Just a nit on this one: I think it would be better to make the textfield fullWidth but set it in the container and then we modify the containers size.

Not necessary tho

apps/root/src/pages/earn/components/all-strategies-table/index.tsx
```diff
@@ -192,30 +203,53 @@ const AllStrategiesTable = () => {
     void fetchStrategies();
   }, []);
 
+  const handleSearchChange = (newValue: string) => {
+    setPage(0);
+    setSearch(newValue);
+  };
+
+  const filteredStrategies = React.useMemo<Strategy[]>(
```
FiboApe, line 211: btw remember that search WILL trigger a call to the sdk to fetch more vaults probably, maybe we should work on the filters + pagination first, so we can have that already done and then when we implement the search is a bit more straightforward

th0rOdinson: Agree, will resume this next week then!

apps/root/src/pages/earn/components/all-strategies-table/components/toolbar/index.tsx
```diff
@@ -1,21 +1,53 @@
...
+          size="small"
+          placeholder={intl.formatMessage(
+            defineMessage({
+              defaultMessage: 'Search by Vault, Chain, Assets, Guardian or Yield Type',
+              description: 'allStrategiesSearch',
+            })
+          )}
+          value={search}
+          onChange={(evt: React.ChangeEvent<HTMLTextAreaElement | HTMLInputElement>) =>
+            setSearch(evt.currentTarget.value)
```
FiboApe, line 32: I'm thinking if we add this to the redux state, we wont need to:
1. Filter strategies on the parent component
2. Reduce prop drilling for search parameters.

What do you think?

th0rOdinson: Like it very much sir!

apps/root/src/pages/earn/components/all-strategies-table/components/toolbar/index.tsx
```diff
@@ -3,16 +3,18 @@ import { useThemeMode } from '@state/config/hooks';
 import { FormattedMessage, defineMessage, useIntl } from 'react-intl';
 import { ContainerBox, InputAdornment, SearchIcon, TextField, Typography, colors } from 'ui-library';
 import TableFilters from '../filters';
+import { useAllStrategiesFilters } from '@state/all-strategies-filters/hooks';
 
 interface AllStrategiesTableToolbarProps {
   isLoading: boolean;
-  search: string;
-  setSearch: (newValue: string) => void;
+  setSearch: (search: string) => void;
```
FiboApe: no need to prop drill this, you can just use it below

FiboApe (approved):

PR #840, opened 2024-06-07
refactor(*): migrate to viem 2.x and wagmi 2.x

FiboApe:


apps/root/src/hooks/useSwapOptions.ts
```diff
@@ -104,7 +104,21 @@ function useSwapOptions(
...
+                setState({ result: promiseResult, error: undefined, isLoading: false });
+              } else {
+                await aggregatorService
+                  .buildSwapOptions({ options: promiseResult, recipient: debouncedAccount as Address })
+                  .then((builtResponse) => {
+                    const newResults = promiseResult.map((option) => ({
+                      ...option,
+                      tx: builtResponse[option.swapper.id],
+                    }));
+                    setState({ result: newResults, error: undefined, isLoading: false });
```
th0rOdinson, line 118: I think we should exclude from `promiseResult` all quotes that dont are included in builtResponse (no tx)

PR #852, opened 2024-06-17
feat(earn): filter for all strategies

th0rOdinson:


packages/common-types/src/earn.ts
```diff
@@ -1,36 +1,38 @@
 import { Address } from 'viem';
-import { NetworkStruct, TokenWithIcon } from '.';
+import { NetworkStruct, TokenListId, TokenWithIcon } from '.';
 
-export type ApiStrategy = {
+export type SdkStrategy = {
```
FiboApe, line 4: Can't we just use now the ones exported by the SDK?

apps/root/src/common/utils/earn/parsing.tsx
```diff
@@ -63,24 +61,42 @@ export const parseAllStrategies = ({
...
     return {
       id: strategy.id,
-      asset: earnApiTokenToTokenWithIcon(strategy.asset, strategy.chainId, tokenList),
-      rewards: Object.values(strategy.rewards).map((reward) => ({
-        token: earnApiTokenToTokenWithIcon(reward.token, strategy.chainId, tokenList),
-        apy: reward.apy,
-      })),
+      asset: {
+        ...assetToken,
+        icon: <TokenIcon token={assetToken} />,
```
FiboApe, line 80: I'm not sure we should start sending a lot of react components through parsing, not really sold on the idea, I would try to keep the parsing from address/asset into React component as far along the component tree as we can. Since we cannot store react components for example

apps/root/src/hooks/earn/useFilteredStrategies.ts
```diff
@@ -0,0 +1,38 @@
...
+      const isFarmMatch = filtersApplied.farms.length === 0 || filtersApplied.farms.includes(strategy.farm.id);
+
+      const isGuardiansMatch =
+        filtersApplied.guardians.length === 0 || filtersApplied.guardians.includes(strategy.guardian?.id || '');
+
+      return isAssetMatch && isRewardMatch && isNetworkMatch && isYieldTypeMatch && isFarmMatch && isGuardiansMatch;
+    });
+  }, [strategies, filtersApplied]);
+
+  return { filteredStrategies, isLoadingAllStrategies };
```
FiboApe, line 37: Oh lets memo this, since if not its going to create a new object on every re-render

apps/root/src/state/all-strategies-filters/actions.ts
```diff
@@ -0,0 +1,14 @@
+import { createAction } from '@reduxjs/toolkit';
+import { ChainId, StrategyYieldType, Token } from 'common-types';
+
+export const setAssetFilter = createAction<Token[]>('allStrategiesFilters/setAssetFilter');
+
+export const setNetworkFilter = createAction<ChainId[]>('allStrategiesFilters/setNetworkFilter');
+
+export const setRewardFilter = createAction<Token[]>('allStrategiesFilters/setRewardFilter');
+
+export const setFarmFilter = createAction<string[]>('allStrategiesFilters/setFarmFilter');
```
FiboApe, line 10: oh we could create a FarmId type or GuardianId type also, so we start moving away from primitives

apps/root/src/pages/earn/components/all-strategies-table/index.tsx
```diff
@@ -147,7 +149,7 @@ const AllStrategiesTableBodySkeleton = () => (
 );
 
 const createRow = (strategy: Strategy) => (
-  <TableRow>
+  <TableRow key={`${strategy.id}-${strategy.guardian?.id}`}>
```
FiboApe, line 152: stategy id is already going to be unique :) since it refers to the farm with the specific guardian

apps/root/src/pages/earn/components/all-strategies-table/components/toolbar/index.tsx
```diff
@@ -0,0 +1,23 @@
...
+interface AllStrategiesTableToolbarProps {
+  isLoading: boolean;
+}
+
+const AllStrategiesTableToolbar = ({ isLoading }: AllStrategiesTableToolbarProps) => {
+  const themeMode = useThemeMode();
+  return (
+    <ContainerBox justifyContent="space-between" alignItems="end">
+      <Typography variant="h4Bold" color={colors[themeMode].typography.typo1}>
+        <FormattedMessage description={'allVaults'} defaultMessage={'All Vaults'} />
```
FiboApe, line 16: copilot did his thing here 👁️

apps/root/src/pages/earn/components/all-strategies-table/components/filters/index.tsx
```diff
@@ -0,0 +1,325 @@
...
+      handleFilterChange: (filter) => dispatch(setGuardianFilter(filter)),
+    };
+
+    return [networksFilter, assetFilter, rewardsFilter, farmsFilter, yieldTypesFilter, guardiansFilter];
+  }, [intl, strategiesFilters, strategiesParameters]);
+
+  return (
+    <>
+      <Button onClick={handleOpen} disabled={isLoading} variant="outlined" endIcon={<KeyboardArrowDownIcon />}>
+        <FormattedMessage defaultMessage="Filters" description="all-strategies-table.filters" />
```
FiboApe, line 290: can you check with emi if this works for him?

apps/root/src/pages/earn/components/all-strategies-table/components/filters/index.tsx
```diff
@@ -0,0 +1,325 @@
...
+      </Button>
+      <Popover
```
FiboApe, line 292: Seems like we overly use this, maybe it would be good to have it componentized with overridable params?

apps/root/src/pages/earn/components/all-strategies-table/components/filters/index.tsx
```diff
@@ -0,0 +1,325 @@
...
+          horizontal: 'left',
+        }}
+        anchorEl={anchorEl}
+        id={id}
+        open={!isLoading && open}
+        onClose={handleClose}
+        disableAutoFocus
+      >
+        {filterItems.map((filters, index) => (
+          <Filter<FilterTypes>
```
FiboApe, line 308: do you actually need to send FilterTypes here?

apps/root/src/pages/earn/components/all-strategies-table/components/filters/index.tsx
```diff
@@ -0,0 +1,325 @@
...
+
+  const handleOpen = (event: React.MouseEvent<HTMLButtonElement>) => {
+    setAnchorEl(event.currentTarget);
+  };
+
+  const open = Boolean(anchorEl);
+  const id = open ? 'strategies-filters-popover' : undefined;
+
+  const filterItems = React.useMemo<FilterControl<FilterTypes>[]>(() => {
+    const assetFilter: FilterControl<Token> = {
```
FiboApe, line 177: seems like a lot of this are just duplicated code, can we maybe make a function to generate those?

apps/root/src/hooks/earn/useStrategiesParameters.ts
```diff
@@ -0,0 +1,58 @@
...
+
+    return {
+      farms: Object.values(strategiesParameters.farms),
+      networks: Object.values(strategiesParameters.networks),
+      assets: removeEquivalentFromTokensArray(assets),
+      rewards: removeEquivalentFromTokensArray(rewards),
+      yieldTypes,
+      guardians: Object.values(strategiesParameters.guardians),
+    };
+  }, [strategiesParameters, tokenList, intl, isLoadingAllStrategies]);
```
FiboApe, line 57: you are not actually using isLoadingAllStrategies here :)

apps/root/src/pages/earn/components/all-strategies-table/components/filters/index.tsx
th0rOdinson: Sure, we agreed to use a hierarchical structure from now on. In this case: `earn.all-strategies-table.filters`

apps/root/src/common/utils/earn/parsing.tsx
th0rOdinson: tsx => tx, got it! I remember now the issue you faced with redux with a similar parsing function

packages/common-types/src/earn.ts
th0rOdinson: I just pinged out BE team, and it seems they are far from implementing this in the SDK. They suggest we replicate the types in the meantime.

apps/root/src/pages/earn/components/all-strategies-table/components/filters/index.tsx
th0rOdinson: I have created a helper function for this, but I believe some repetition is necessary due to the specific requirements of each option. Let me know if you have any other approaches in mind!

apps/root/src/pages/earn/components/all-strategies-table/components/filters/index.tsx
th0rOdinson: Actually this is the second time we are using it. I tried to abstract it but the first implementation has a different button component, so the open/close behaviour should be handled by the consumer of the component, we wouldn't really gain much in my opinion. I have set some styles by default in our theme though! And also remove code from our old footer

apps/root/src/hooks/earn/useStrategiesParameters.ts
th0rOdinson: This issue was caused by a re-render problem where the hook was not updating correctly. The root cause was that the service was providing a nested object. In this case, I chose to use `cloneDeep` because it's a one-time update. For other use cases, we should consider a more performance-efficient solution

packages/common-types/src/earn.ts
FiboApe: Oh that's sad, ok then

apps/root/src/pages/earn/components/all-strategies-table/components/filters/index.tsx
FiboApe: awesome!

apps/root/src/services/earnService.ts
```diff
@@ -54,7 +55,8 @@ export class EarnService extends EventsManager<EarnServiceData> {
   }
 
   set strategiesParameters(strategiesParameters) {
-    this.serviceData = { ...this.serviceData, strategiesParameters };
+    // Usefull for updating nested objects
+    this.serviceData = { ...this.serviceData, strategiesParameters: cloneDeep(strategiesParameters) };
```
FiboApe, line 59: wait why do we actually need to cloneDeep here?, the strategiesParameter is going to be another reference, and we are updating the reference of this.serviceData so it should be a new object triggering the useServiceData

FiboApe (approved):

PR #855, opened 2024-06-18
Revert "Revert balances"

th0rOdinson:
Reverts Balmy-protocol/dca-fe#853


FiboApe (approved):
This is not "Release" tho its just reverts, jusst so we dont get the merge commit/ github description weird with a merge to staging

PR #857, opened 2024-06-18
fix(transfer): sort tokens by usd number and not alphabetical

FiboApe:


apps/root/src/common/components/token-selector/index.tsx
```diff
@@ -150,7 +150,7 @@ const TokenSelector = ({ handleChange, selectedToken }: TokenSelectorProps) => {
       parsedWithCustomTokens,
       [
         ({ balanceUsd, token }) =>
-          formatUsdAmount({ amount: formatUnits(balanceUsd || 0n, token.decimals + 18), intl }),
+          Number(formatUsdAmount({ amount: formatUnits(balanceUsd || 0n, token.decimals + 18), intl })),
```
th0rOdinson, line 153: This can fail if number is comma separated => Number("1,1") = NaN

FiboApe: Oh really good carch

th0rOdinson (approved):

PR #869, opened 2024-06-25
feat(earn): create strategy details frame

th0rOdinson:


apps/root/src/frame/index.tsx
```diff
@@ -144,6 +145,7 @@ const AppFrame = ({ config: { wagmiClient }, initialChain }: AppFrameProps) => {
                                 path="/:chainId/positions/:positionVersion/:positionId"
                                 element={<PositionDetail />}
                               />
+                              <Route path=":chainId/vaults/:strategyGuardianId" element={<StrategyGuardianDetail />} />
```
FiboApe, line 148: I know this might sound like a break in patterns, but I woudl really make this `/earn/:chainId/vaults/:strategyGuardianId`

I hate how the position one ended up in

apps/root/src/pages/strategy-guardian-detail/frame/index.tsx
```diff
@@ -0,0 +1,52 @@
...
+  const trackEvent = useTrackEvent();
+  const intl = useIntl();
+
+  React.useEffect(() => {
+    dispatch(changeRoute(EARN_ROUTE.key));
+    trackEvent('Earn - Visit strategy detail page', { chainId });
+  }, []);
+
+  const onBackToEarnHome = () => {
+    dispatch(changeRoute(DCA_ROUTE.key));
```
FiboApe, line 29: nono, this should go to earn :)

apps/root/src/pages/strategy-guardian-detail/frame/index.tsx
```diff
@@ -0,0 +1,52 @@
...
+    strategyGuardianId: string;
+    chainId: string;
+  }>();
+  const pushToHistory = usePushToHistory();
+  const dispatch = useAppDispatch();
+  const trackEvent = useTrackEvent();
+  const intl = useIntl();
+
+  React.useEffect(() => {
+    dispatch(changeRoute(EARN_ROUTE.key));
```
FiboApe, line 24: This one would actually change based on if the user had an investment there right? Since we would like to highlight the "portfolio" sub item

apps/root/src/frame/index.tsx
th0rOdinson: Me too, haha. I believe Earn can be an opportunity to improve other DCA routes (that can be optimized, aside from the position route). I can create a task for it if you agree

apps/root/src/pages/strategy-guardian-detail/frame/index.tsx
th0rOdinson: So true!

FiboApe (approved):

apps/root/src/frame/index.tsx
FiboApe: Please do :). Because we also have to keep backwards compatibility

PR #875, opened 2024-06-28
fix(dca): transfering position to owned wallet

th0rOdinson:
https://github.com/Balmy-protocol/dca-fe/assets/144699868/aa688aa8-161d-4150-acc4-2f308add034b

### DIFF
--- a/apps/root/src/services/positionService.ts
+++ b/apps/root/src/services/positionService.ts
@@ -1818,7 +1818,17 @@ export default class PositionService extends EventsManager<PositionServiceData> 
       }
       case TransactionTypes.transferPosition: {
         const transferPositionTypeData = transaction.typeData;
-        delete currentPositions[transferPositionTypeData.id];
+        const wallets = this.accountService.getWallets();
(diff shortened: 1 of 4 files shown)


apps/root/src/services/positionService.ts
```diff
@@ -1818,7 +1818,17 @@ export default class PositionService extends EventsManager<PositionServiceData>
...
       case TransactionTypes.transferPosition: {
         const transferPositionTypeData = transaction.typeData;
-        delete currentPositions[transferPositionTypeData.id];
+        const wallets = this.accountService.getWallets();
+        const recipientIsSameUser = wallets.some(
+          (wallet) => wallet.address.toLowerCase() === transaction.typeData.toAddress.toLowerCase()
+        );
+
+        if (recipientIsSameUser) {
+          currentPositions[transferPositionTypeData.id].user = transaction.typeData.toAddress.toLowerCase() as Address;
```
FiboApe, line 1827: We also need to clear the permissions, since those are cleared on a transfer :)

FiboApe (approved):

PR #887, opened 2024-07-04
fix(*): dont await for ens

th0rOdinson:


apps/root/src/services/conctactListService.ts
```diff
@@ -131,8 +131,10 @@ export default class ContactListService extends EventsManager<ContactListService
     }
 
     const wallets = this.accountService.user?.wallets || [];
-    const accountEns = await this.walletService.getManyEns(map(wallets, 'address'));
-    this.accountService.setWalletsEns(accountEns);
+    void this.walletService.getManyEns(map(wallets, 'address')).then((ens) => {
```
FiboApe, line 134: we should actually modify the getManyEns, if one fails we dont want all to fail, and at least to load the ones we can

th0rOdinson: Sure, I'll handle any errors on that side. This issue was not caused by that though, we have a timeout for this `initializeAliasesAndContacts` that was meant to be related to the api call, but the ENS were delaying it all

FiboApe (approved):

PR #888, opened 2024-07-04
feat(earn): add getUserVaults and dynamic shift of vault details

FiboApe:


apps/root/src/common/utils/earn/parsing.tsx
```diff
@@ -86,3 +88,48 @@ export const parseAllStrategies = ({
       ...rest,
     };
   });
+
+export const parseUserStrategies = ({
```
th0rOdinson, line 92: Our data flow is just ✨

apps/root/src/hooks/useNetWorth.ts
```diff
@@ -95,8 +98,35 @@ const useNetWorth = ({ walletSelector, chainId }: NetWorthProps) => {
     0
   );
 
+  // EARN
+  let earnAssetsTotalValue = 0;
+  let filteredUserStrategies = userStrategies;
+  if (walletSelector !== ALL_WALLETS) {
+    if (chainId && walletAddressToEvaluate) {
```
th0rOdinson, line 105: This condition can be simplified with the condition above right? And maybe in DCA too 😄

packages/common-types/src/earn.ts
```diff
@@ -97,19 +187,50 @@ export type BaseDetailedStrategy = BaseStrategy & {
...
 
-export type GuardianId = string;
-export type FarmId = string;
-export type StrategyId = string;
+export type BaseEarnPosition = {
+  id: SdkEarnPositionId;
+  createdAt: Timestamp;
+  owner: Address;
+  permissions: EarnPermissions;
+  strategy: Strategy;
```
th0rOdinson, line 207: I'm thinking if we could have one strategy object. Like we did for details, merging it into base-strategy, what if we also merge position-data into the allStrategies? We would avoid having duplicated data, and it would be easier to check if a strategy is a position or not, without calling the two lists (like in earn home, or in vault details). Not sure if you have already explored this option, let me know

apps/root/src/services/sdkService.ts
```diff
@@ -403,19 +410,62 @@ export default class SdkService {
...
+    //   ids: [positionStrategyId],
+    //   includeHistory: true,
+    //   includeHistoricalbalancesFrom: THREE_MONTHS * 1000,
+    // })
+    // getPositionsById(_: {
+    //   ids: ArrayOneOrMore<{ chainId: ChainId; positionId: BigIntish }>;
+    //   includeHistory?: boolean;
+    //   includeHistoricalBalancesFrom?: Timestamp;
+    //   config?: { timeout: TimeString };
+    // }): Promise<Record<ChainId, EarnPosition[]>>;
```
th0rOdinson, line 461: good idea having this as a comment

apps/root/src/common/components/wallet-selector/index.tsx
```diff
@@ -189,6 +191,10 @@ const WalletSelector = ({ options, size = 'small' }: WalletSelectorProps) => {
           description: ApiErrorKeys.DCA_POSITIONS,
         }).then(() => void dispatch(processConfirmedTransactions()));
 
+        await timeoutPromise(earnService.fetchUserStrategies(), TimeoutPromises.COMMON, {
```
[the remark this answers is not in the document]
FiboApe: I would actually like to show an error in case we haven't loaded them in time, since it means something is off

apps/root/src/hooks/useNetWorth.ts
FiboApe: I mean yeah but we still ahve that other condition and it is in the scope of the `1walletSelector !== ALL_WALLETS`

apps/root/src/common/components/wallet-selector/index.tsx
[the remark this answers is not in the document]
th0rOdinson: But should be delay fetchInitialBalances because of this?

apps/root/src/pages/strategy-guardian-detail/vault-data/components/data-about/index.tsx
```diff
@@ -100,27 +148,73 @@ const DataAbout = ({ strategy }: DataAboutProps) => {
...
+                    <FormattedMessage
+                      defaultMessage="Balmy fees"
+                      description="earn.strategy-details.vault-about.balmy-fee"
+                    />
+                  }
+                  content={
+                    <FeeContainer
+                      isLoading={isLoading}
+                      intl={intl}
+                      fees={[{ percentage: 0.2, type: FeeType.performance }]}
```
th0rOdinson, line 178: Could we mock this `0.2`?

th0rOdinson (approved):

PR #890, opened 2024-07-05
feat(mixpanel): add share on x metric

th0rOdinson:


FiboApe (approved):
this doesnt break the event propagation and the link still works right?

th0rOdinson:
> this doesnt break the event propagation and the link still works right?

No it doesn't! I've tested it and it works fine 😄

PR #894, opened 2024-07-09
Webpack improvements

th0rOdinson:
### Before
Screenshot 2024-07-07 at 22 56 36

### New implementation
Screenshot 2024-07-09 at 2 10 40 PM

- Updated ts-loader version
- Implemented a new plugin in order to run ts checks in a parallel process
- Skip minifying in development
- Added a new plugin to measure better webpack performance

### Result
- (+) Reduced build time by 60 secs aproximately (in both development and production)
- (+) `NODE_OPTIONS=--max_old_space_size=8192` can be now safely removed 
- (-) Increased total .js assets from 52 to 58 MiB (only in develpoment)


apps/root/webpack.common.js
```diff
@@ -91,6 +45,7 @@ module.exports = {
       'process.env.ENABLED_TRANSLATIONS': JSON.stringify(process.env.ENABLED_TRANSLATIONS),
     }),
     new WebpackBar(),
+    new ForkTsCheckerWebpackPlugin(),
```
FiboApe, line 48: Screenshot 2024-07-10 at 10 49 11 AM
Should we modify this so it can use a bit more memory?

apps/root/webpack.dev.js
```diff
@@ -1,15 +1,25 @@
...
+    devServer: {
+      port: 3000,
+      historyApiFallback: true,
+      hot: true,
+      client: {
+        overlay: false,
+      },
+    },
+    snapshot: {
+      managedPaths: [/^(.+?[\\/]node_modules[\\/](?!(ui-library|common-types))(@.+?[\\/])?.+?)[\\/]/],
```
FiboApe, line 22: what's this for? So it can pick up changes on the other repos?

apps/root/webpack.prod.js
```diff
@@ -12,4 +13,50 @@ module.exports = merge(common, {
       'process.env.ARCX_KEY': JSON.stringify(process.env.ARCX_KEY),
     }),
   ],
+  optimization: {
+    minimize: true,
+    minimizer: [
+      // This is only used in production mode
+      new TerserPlugin({
```
FiboApe, line 20: Have you researched any options to make the prod bundle size smaller?

apps/root/webpack.dev.js
th0rOdinson: Exactly, I followed [this documentation](https://webpack.js.org/configuration/other-options/#managedpaths)
Hot reload stopped working as it did before; I think this might be because it now builds the frontend before `build:packages` finishes. This configuration suits our use case

FiboApe: Ok so having a watch on those packages still trigger a re-render right?

apps/root/webpack.prod.js
th0rOdinson: I explored some bundle analyzers and found that a significant portion (around 22 MB) comes from node_modules. To improve this, we should define a new spike in my opinion.
Regarding our app itself, I didn't find any significant areas for improvement during this spike. There are other minimizer plugins available, but I didn't spend time exploring them. 

Something to highlight: Our emojis folder only is 545 Kb, we should keep this in mind for the future

This is our current bundle breakdown for the app, and it's close to what we can expect from it:
Screenshot 2024-07-10 at 2 03 04 PM

apps/root/webpack.dev.js
th0rOdinson: Yes, the server restarts on any changes from this directory. Running build:packages triggers the server a couple of times, due to the deletion and creation of distribution files.
I have updated it in the new commit, using a more modern property (`unmanagedPaths`). This managedPaths output something extra in the console, this is cleaner and more maintainable for any new package

yarn.lock
```diff
@@ -16824,7 +16829,7 @@ webpack-virtual-modules@^0.6.1:
   resolved "https://registry.yarnpkg.com/webpack-virtual-modules/-/webpack-virtual-modules-0.6.2.tgz#057faa9065c8acf48f24cb57ac0e77739ab9a7e8"
   integrity sha512-66/V2i5hQanC51vBQKPH4aI8NMAcBW59FVBs+rC7eGHupMyfn34q7rZIE+ETlJ+XTevqfUhVVBgSUNSW2flEUQ==
 
-webpack@5, webpack@^5.37.1, webpack@^5.88.2:
+webpack@5, webpack@^5.88.2:
```
FiboApe, line 16832: Seems like all webpack dependencies could be merged together. My recomendation is delete this one and the bottom one, and run yarn install, it should create the same dependency for all the versions

FiboApe (approved):
Please create the spike for optimizing the prod bundle

PR #898, opened 2024-07-09
fix(transactionSteps): special label for Unlimited approval

th0rOdinson:
Screenshot 2024-07-09 at 2 53 06 PM

### DIFF
--- a/apps/root/src/common/components/transaction-steps/index.tsx
+++ b/apps/root/src/common/components/transaction-steps/index.tsx
@@ -22,6 +22,7 @@ import {
   TransactionActionApproveTokenSignSwapType,
   SetStateCallback,
   TransactionApplicationIdentifier,
+  TransactionEventTypes,
 } from '@types';
 import {
   TRANSACTION_ACTION_SWAP,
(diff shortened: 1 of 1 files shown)


apps/root/src/common/components/transaction-steps/index.tsx
```diff
@@ -436,14 +436,24 @@ const buildApproveTokenItem = ({
...
                   label={
-                    <FormattedMessage
-                      description="approveAmount"
-                      defaultMessage="Approve {amount} {symbol}"
-                      values={{
-                        symbol: token.symbol,
-                        amount: formatCurrencyAmount({ amount, token, sigFigs: 4, intl }),
-                      }}
-                    />
+                    isPermit2Enabled ? (
```
FiboApe, line 439: This should only be based on the approved amount, don't we have that data?

th0rOdinson: We always have the max amount to approve, based on the quotes received. We dont have the `maxUint256` here. If permit2Enabled, we don't send an amount and the walletService submits unlimited

I know we are handling this different from txEvents, but I don't see how the user could have permit2Enabled, and don't submit an unlimited approval

FiboApe: This is as of now, if we later want to enable users to exact approve permit2 contracts this will break. 

Just as a note, we used to allow users to exact approve the permit2 contract. Nothings says we will enable it again at some point, lets plan for the future :)

FiboApe (approved):

PR #900, opened 2024-07-11
feat(earn): order by all strategies column

th0rOdinson:


apps/root/src/pages/earn/frame/index.tsx
```diff
@@ -4,23 +4,41 @@ import EarnFAQ from '../components/faq';
...
+
   return (
     <StyledNonFormContainer flexDirection="column" flexWrap="nowrap">
       <ContainerBox flexDirection="column" gap={32}>
         <ContainerBox flexDirection="column" gap={20}>
           <EarnWizard />
           <ContainerBox flex="1">
-            <AllStrategiesTable />
+            <AllStrategiesTable
+              defaultOrderBy={StrategyColumnKeys.TVL}
```
FiboApe, line 38: why do we need to pass this as a parameter?

apps/root/src/pages/earn/frame/index.tsx
```diff
@@ -4,23 +4,41 @@ import EarnFAQ from '../components/faq';
...
+              hasFetchedAllStrategies={hasFetchedAllStrategies}
```
FiboApe, line 39: this can be consumed directly on the component, thus making us not re-render everything in this page when that changes

apps/root/src/pages/earn/frame/index.tsx
```diff
@@ -4,23 +4,41 @@ import EarnFAQ from '../components/faq';
...
 
   React.useEffect(() => {
     dispatch(changeRoute(EARN_ROUTE.key));
   }, []);
 
+  React.useEffect(() => {
+    const fetchStrategies = async () => {
+      await earnService.fetchAllStrategies();
+    };
+    if (!hasFetchedAllStrategies) {
```
FiboApe, line 26: Related to the comment below, same here, if we've got this on this frame, hasFetchedAllStrategies will trigger a re-render in all child components, maybe the check only needs to go into the fetchAllStrategies method of the earnService

apps/root/src/pages/earn/components/strategies-table/index.tsx
```diff
@@ -0,0 +1,267 @@
...
+    const isAsc = orderBy === property && order === 'asc';
+    setOrder(isAsc ? 'desc' : 'asc');
+    setOrderBy(property);
+  };
+
+  return (
+    <ContainerBox flexDirection="column" gap={5} flex={1}>
+      <AllStrategiesTableToolbar isLoading={!hasFetchedAllStrategies} handleSearchChange={handleSearchChange} />
+      <StyledBackgroundPaper variant="outlined">
+        <TableContainer component={Paper} elevation={0}>
```
FiboApe, line 234: so a paper inside a paper, cant we just send StyledBackgroundPaper to this tableContainer component?

apps/root/src/pages/earn/components/strategies-table/index.tsx
```diff
@@ -0,0 +1,267 @@
...
+const AllStrategiesTable = ({
+  strategies,
+  hasFetchedAllStrategies,
+  defaultOrderBy,
+}: {
+  strategies: TableStrategy[];
+  hasFetchedAllStrategies: boolean;
+  defaultOrderBy: StrategyColumnKeys;
+}) => {
+  const [order, setOrder] = React.useState<Order>('desc');
```
FiboApe, line 190: If we keep this on redux, the sorting will persist through the user changing pages/tabs, which is pretty cool if you ask me

apps/root/src/pages/earn/components/strategies-table/index.tsx
```diff
@@ -0,0 +1,267 @@
...
+  );
+
+  const handleSearchChange = debounce((newValue: string) => {
+    setPage(0);
+    dispatch(setSearch(newValue));
+  }, 500);
+
+  const visibleRows = React.useMemo<TableStrategy[]>(
+    () =>
+      strategies
```
FiboApe, line 214: what if useFilteredStrategies which already takes in mind the filters, also takes in mind the order, and we just handle that in one point

apps/root/src/pages/earn/components/strategies-table/index.tsx
```diff
@@ -0,0 +1,267 @@
...
+        .slice()
+        .sort(getComparator(order, orderBy))
+        .slice(page * ROWS_PER_PAGE, page * ROWS_PER_PAGE + ROWS_PER_PAGE),
+    [page, strategies, order, orderBy]
+  );
+
+  // Keeps the table height consistent
+  const emptyRows = createEmptyRows(ROWS_PER_PAGE - visibleRows.length);
+
+  const handleRequestSort = (property: StrategyColumnKeys) => {
```
FiboApe, line 224: if we have this on redux, we could not drill prop a lot of this stuff :D

apps/root/src/pages/earn/components/strategies-table/index.tsx
th0rOdinson: Removing that paper got involved adding some styles to our tables in general. Now it looks much better in my opinion!

apps/root/src/pages/earn/components/strategies-table/index.tsx
th0rOdinson: Like it!

apps/root/src/pages/earn/frame/index.tsx
th0rOdinson: About this 3 comments, I based a lot of this PR into making this component as flexible as possible! This 3 props would change from the user strategies table.
About the flag itself, I think it's necessary at this point as an isLoading state. We could move it to the service though, but this useEffect check would be still here

apps/root/src/pages/earn/components/strategies-table/index.tsx
th0rOdinson: A little prop drilling was needed for keeping this flexible :(. Let me know if you have any other approach to take into count!

apps/root/src/pages/earn/frame/index.tsx
```diff
@@ -4,23 +4,51 @@ import EarnFAQ from '../components/faq';
...
       <ContainerBox flexDirection="column" gap={32}>
         <ContainerBox flexDirection="column" gap={20}>
           <EarnWizard />
           <ContainerBox flex="1">
-            <AllStrategiesTable />
+            <AllStrategiesTable
+              isLoading={!hasFetchedAllStrategies}
+              strategies={strategies}
+              onSortChange={onSortChange}
+              orderBy={orderBy}
```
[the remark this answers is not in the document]
th0rOdinson: None, the table just need to know who is calling 😄 
Did some changes related to this now!

apps/root/src/pages/earn/components/strategies-table/index.tsx
```diff
@@ -55,24 +58,29 @@ const StyledNavContainer = styled(ContainerBox)`
...
-  onRequestSort,
+  variant,
 }: {
   columns: StrategyColumnConfig[];
-  order: ColumnOrder;
-  orderBy?: StrategyColumnKeys;
-  onRequestSort: (property: StrategyColumnKeys) => void;
+  variant: StrategiesTableVariants;
 }) => {
+  const dispatch = useDispatch();
```
FiboApe, line 66: oh this should be useAppDispatch, not useDispatch directly from redux

FiboApe (commented):
Really nice! Just one itty bitty comment

FiboApe (approved):

PR #903, opened 2024-07-12
feat(*): improve routes

th0rOdinson:


apps/root/src/constants/routes.tsx
```diff
@@ -35,7 +35,7 @@ export const EARN_ROUTE = {
   icon: <MoneyAddIcon />,
 };
 export const EARN_PORTFOLIO = {
-  label: defineMessage({ description: 'earnPortfolio', defaultMessage: 'Portfolio' }),
-  key: 'earnPortfolio',
+  label: defineMessage({ description: 'navigation.earn.portfolio', defaultMessage: 'Portfolio' }),
+  key: 'earn-portfolio',
```
FiboApe, line 39: this should actually be /earn/portfolio right?, so to handle it like we do for dca

th0rOdinson: Like it

apps/root/src/common/components/redirect-old-route/index.tsx
```diff
@@ -0,0 +1,40 @@
+import React from 'react';
+import useTrackEvent from '@hooks/useTrackEvent';
+import { Navigate, Params, useParams } from 'react-router-dom';
+
+export interface RedirectProps {
+  to: string;
+  oldRoute: string;
+}
+
+const updateTo = (to: string, params: Readonly<Params<string>>) => {
```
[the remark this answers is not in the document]
th0rOdinson: Sure, this approach is suggested by `react-router-dom`, but they don't specify how to handle dynamic parameters :D 
Comments added!

FiboApe (approved):

PR #904, opened 2024-07-15
feat(earn): add points to graph for deposits and withdraws

FiboApe:


apps/root/src/services/sdkService.ts
```diff
@@ -472,7 +472,7 @@ export default class SdkService {
     const mockedEarnPositions = new Promise<SdkEarnPosition>((resolve) => {
       setTimeout(() => {
         resolve({ ...sdkDetailedEarnPositionMock, id: positionStrategyId });
-      }, 1000);
+      }, 0);
```
th0rOdinson, line 475: Why? :( We can keep loading states

apps/root/src/services/earnService.ts
```diff
@@ -324,6 +337,24 @@ export class EarnService extends EventsManager<EarnServiceData> {
...
+  }
+
+  async fetchMultipleEarnPositionsFromStrategy(strategyId: StrategyId) {
+    const userStrategies = this.userStrategies;
+
+    const positionsToFetch = userStrategies.filter((strat) => strat.strategy === strategyId);
+
+    const promises = positionsToFetch.map((position) => this.fetchUserStrategy(position.id, false));
+
+    const results = compact(await Promise.all(promises));
```
th0rOdinson, line 354: We should catch any error, so successful promises can continue

apps/root/src/pages/strategy-guardian-detail/vault-data/components/data-historical-rate/index.tsx
```diff
@@ -61,18 +67,224 @@ interface DataHistoricalRateProps {
...
+
+  if (!firstPayload || !firstPayload.payload) {
+    return null;
+  }
+
+  const actions = firstPayload.payload.actions;
+
+  return (
+    <StyledTooltipContainer>
+      {actions.map(({ user, value, type, timestamp }, key) => (
```
th0rOdinson, line 143: Tooltip is rendered as an empty square when there are no actions

apps/root/src/pages/strategy-guardian-detail/frame/index.tsx
```diff
@@ -72,6 +74,16 @@ const StrategyDetailFrame = () => {
...
 
+  React.useEffect(() => {
+    if (strategyGuardianId) {
+      try {
+        void earnService.fetchMultipleEarnPositionsFromStrategy(strategyGuardianId);
+      } catch (error) {
+        console.error('Failed to fetch detailed strategy', chainId, strategyGuardianId, error);
+      }
+    }
+  }, [strategyGuardianId, earnPositions]);
```
th0rOdinson, line 85: Why should this be updated whenever a user position changes? Or is this just to trigger the `needsToUpdateUserStrategy` check?

th0rOdinson (commented):
Really nice!

apps/root/src/services/sdkService.ts
FiboApe: Oh shit yeah, was debugging and forgot to unstage this changes

apps/root/src/services/earnService.ts
FiboApe: Oh good catch

apps/root/src/pages/strategy-guardian-detail/frame/index.tsx
FiboApe: This is to be triggered assuming earnpositions are already on the service and to update their data, since we are going to need a lot of them and didn't want us to be overriding stuff on the go

PR #909, opened 2024-07-15
feat(*): add custom sync of user settings

FiboApe:


apps/root/src/state/index.ts
```diff
@@ -90,8 +93,23 @@ export interface ExtraArgument {
...
 }
 
-const createStore = (web3Service: Web3Service) =>
-  configureStore({
+const persistedSavedActions = [...AggregatorSettingsSavedActions, ...ConfigSavedActions, ...CustomTokensSavedActions];
+
+const createStore = (web3Service: Web3Service) => {
+  const listenerMiddleWare = createListenerMiddleware({ extra: { web3Service } });
+
+  listenerMiddleWare.startListening({
```
th0rOdinson, line 101: Fantastic

apps/root/src/services/accountService.ts
```diff
@@ -341,6 +342,12 @@ export default class AccountService extends EventsManager<AccountServiceData> {
         signature: storedSignature,
       };
 
+      const config = accounts[0].config;
+
+      if (config) {
+        this.web3Service.onUpdateConfig(config);
```
th0rOdinson, line 348: Just to check, isn't empty state an empty JSON? (Current status of all users)

apps/root/src/services/meanApiService.ts
```diff
@@ -29,6 +29,11 @@ import { CLAIM_ABIS } from '@constants/campaigns';
 // MOCKS
 import { getProtocolToken, getWrappedProtocolToken } from '@common/mocks/tokens';
 import { Address, PublicClient, getContract } from 'viem';
+import { SavedCustomConfig } from '@state/base-types';
+
+type AccountWithConfig = Account & {
```
th0rOdinson, line 34: We can move this to a types file

apps/root/src/state/aggregator-settings/reducer.ts
```diff
@@ -56,6 +58,32 @@ export default createReducer(initialState, (builder) => {
     .addCase(setPermit2, (state, { payload }) => {
       state.isPermit2Enabled = payload;
     })
+    .addCase(hydrateAggregatorSettings, (state, { payload }) => {
+      if (!isUndefined(payload.confettiParticleCount)) {
```
th0rOdinson, line 62: Isn't confetti now fixed for everyone? Or you keep this for old users

apps/root/src/services/accountService.ts
```diff
@@ -341,6 +342,12 @@ export default class AccountService extends EventsManager<AccountServiceData> {
...
+      }
```
th0rOdinson, line 349: What about current config? We would only be storing new changes applied to the store?

apps/root/src/services/meanApiService.ts
FiboApe: Ish, it actually uses a type that is explicitely exclusive to the app, so we don't have access to RootState from the types package :(

apps/root/src/state/aggregator-settings/reducer.ts
FiboApe: Old users + if we later change it and make it configurable again, its already there

th0rOdinson (approved):

PR #910, opened 2024-07-17
feat(tokenProfile): Create frame

th0rOdinson:
Created token profile frame, and some refactor+fixes around useToken hook and custom tokens


apps/root/src/hooks/useToken.ts
```diff
@@ -1,31 +1,64 @@
...
-  }
-
-  if (tokenAddress.toLowerCase() === PROTOCOL_TOKEN_ADDRESS && chainId) return getProtocolToken(chainId);
+interface UseTokenProps {
+  tokenAddress?: string;
+  checkForSymbol?: boolean;
+  filterForDca?: boolean;
+  chainId?: number;
+}
+function useToken({
```
FiboApe, line 16: Love the changes to useToken!

apps/root/src/pages/aggregator/swap-container/index.tsx
```diff
@@ -30,9 +30,33 @@ const SwapContainer = () => {
   const currentNetwork = useSelectedNetwork();
   const isPermit2Enabled = useIsPermit2Enabled(currentNetwork.chainId);
   const { from: fromParam, to: toParam, chainId } = useParams<{ from: string; to: string; chainId: string }>();
-  const fromParamToken = useToken(fromParam, true, false, Number(chainId) || undefined);
-  const toParamToken = useToken(toParam, true, false, Number(chainId) || undefined);
+
+  const sdkMappedNetworks = useSdkMappedChains();
+  const mappedNetworks = React.useMemo(
```
FiboApe, line 35: cant we just use 1 memo for all of this?

apps/root/src/pages/dca/create-position/index.tsx
```diff
@@ -21,8 +21,18 @@ const SwapContainer = ({ handleChangeNetwork }: SwapContainerProps) => {
   const dispatch = useAppDispatch();
   const currentNetwork = useCurrentNetwork();
   const { from: fromParam, to: toParam, chainId } = useParams<{ from: string; to: string; chainId: string }>();
-  const fromParamToken = useToken(fromParam, true, false, Number(chainId));
-  const toParamToken = useToken(toParam, true, false, Number(chainId));
+  const fromParamToken = useToken({
+    chainId: currentNetwork.chainId,
```
FiboApe, line 25: shouldn't this be the Number(chainId)?

apps/root/src/pages/token-profile/frame/index.tsx
```diff
@@ -0,0 +1,83 @@
...
+    <StyledNonFormContainer>
+      <Grid container direction="column" alignItems="stretch" spacing={8}>
+        <Grid item xs={12}>
+          <ContainerBox flexDirection="column" gap={6}>
+            <BackControl
+              onClick={handleGoBack}
+              label={intl.formatMessage(defineMessage({ defaultMessage: 'Back', description: 'back' }))}
+            />
+            {token.symbol}
+            {/* // Header: BLY-2746 */}
```
FiboApe, line 52: Oh love having the ticket numbers here, good practice

apps/root/src/pages/transfer/components/transfer-form/index.tsx
```diff
@@ -99,7 +99,12 @@ const TransferForm = () => {
   const trackEvent = useTrackEvent();
   const { token: selectedToken, recipient, amount } = useTransferState();
   const selectedNetwork = useSelectedNetwork();
-  const tokenParam = useToken(tokenParamAddress, false, false, Number(chainIdParam));
+  const tokenParam = useToken({
+    tokenAddress: tokenParamAddress,
+    checkForSymbol: true,
+    filterForDca: false,
+    chainId: selectedNetwork.chainId,
```
FiboApe, line 106: should this be Number(chainIdParam)?

apps/root/src/hooks/useToken.ts
```diff
@@ -1,31 +1,64 @@
...
+    // Try exact match first (chainId + Address)
+    if (chainId && isAddress(tokenAddress)) {
+      const tokenListId = `${chainId}-${tokenAddress}` as TokenListId;
+      const foundByTokenListId = chainId && tokenList[tokenListId];
 
-  const foundToken = tokenList[key];
+      if (foundByTokenListId) {
+        return foundByTokenListId;
+      } else {
+        void dispatch(
```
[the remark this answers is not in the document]
th0rOdinson: Yes, but also only if chainId and address was provided. I think that checking only address/symbol it's a separate case, where custom tokens can't be involved, that's why I included as a fallback when having a correct chainId, and a correct address

apps/root/src/pages/dca/create-position/index.tsx
th0rOdinson: We were not recognizing network names, this fixes that.
When the chainId param is `optimism`, its recognized and dispatched to the store. So we will admit urls that have /ethereum/custom-address

apps/root/src/pages/transfer/components/transfer-form/index.tsx
th0rOdinson: Same response from above!

apps/root/src/pages/dca/create-position/index.tsx
FiboApe: There is no place here where it actually updates the currentNetwork (which btw refers to the wallet connected network)

apps/root/src/pages/transfer/components/transfer-form/index.tsx
FiboApe: This would then work just on a second re-render right? cant we parse the network before so its already set here, and then do the dispatches?

apps/root/src/pages/dca/create-position/index.tsx
th0rOdinson: It was the selected network, imported as current network 🫠

apps/root/src/pages/transfer/components/transfer-form/index.tsx
th0rOdinson: Did some changes around this! and also some token-resetting when changing the network

PR #914, opened 2024-07-20
feat: build fetchWalletsAllowances

0xsambugs:
We are now implementing `fetchWalletsAllowances`. It will take a list of wallets +  a list of contracts to check (per chain), and it will return the allowances for those contracts (considering all tokens)

### DIFF
--- a/apps/root/tsconfig.json
+++ b/apps/root/tsconfig.json
@@ -1,6 +1,7 @@
 {
   "extends": "tsconfig/react-app.json",
   "compilerOptions": {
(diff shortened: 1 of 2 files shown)


apps/root/tsconfig.json
```diff
@@ -1,6 +1,7 @@
 {
   "extends": "tsconfig/react-app.json",
   "compilerOptions": {
+    "lib": ["ESNext"],
```
0xsambugs, line 4: This changed added cool features like `.flat`, `.flatMap` and more

PR #918, opened 2024-07-24
feat(earn): strategy timeline

th0rOdinson:


apps/root/src/common/mocks/earn.ts
```diff
@@ -207,6 +208,7 @@ export const sdkDetailedEarnPositionMock: DetailedSdkEarnPosition = {
         hash: '0xhash',
         timestamp: 1720042607,
       },
+      assetPrice: 1,
```
FiboApe, line 211: does this actually come in the sdk call?

apps/root/src/pages/home/components/token-icon-multichain/index.tsx
```diff
@@ -26,14 +26,27 @@ const StyledNetworkLogosContainer = styled.div`
...
   const intl = useIntl();
 
   return (
     <ContainerBox flexDirection="column" gap={1}>
       {balanceTokens.map((chainData, index) => (
         <ContainerBox key={index} gap={1} alignItems="center" justifyContent="start">
-          <TokenIcon token={toToken({ logoURI: getGhTokenListLogoUrl(chainData.token.chainId, 'logo') })} size={3.5} />
+          <TokenIcon
+            token={
+              isTokenBreakdown
```
FiboApe, line 44: what is this for?

apps/root/src/pages/position-detail/components/summary-container/index.tsx
```diff
@@ -53,9 +155,25 @@ const PositionSummaryContainer = ({ position, pendingTransaction, isLoading }: P
             <GraphContainer position={position} isLoading={isLoading} />
           </Grid>
           <Grid item xs={12}>
-            <StyledPaper>
-              <PositionSwaps position={position} isLoading={isLoading} />
-            </StyledPaper>
+            <PositionTimelineControls
```
FiboApe, line 158: could we abstract this in just one component? Since it needs to define a lot of things here

apps/root/src/pages/position-detail/components/summary-container/index.tsx
```diff
@@ -53,9 +155,25 @@ const PositionSummaryContainer = ({ position, pendingTransaction, isLoading }: P
...
+              }
+              renderHeader={(item) =>
+                buildDcaTimelineHeader(
+                  MESSAGE_MAP[item.positionState.action](item.positionState, item.position).title,
+                  item.positionState,
+                  item.position.chainId
+                )
+              }
+              renderIcon={(item) => MESSAGE_MAP[item.positionState.action](item.positionState, item.position).icon}
+              getItemId={(item) => `${item.position.chainId}-${item.positionState.tx.hash}`}
```
FiboApe, line 171: this could be defined inside the item?

apps/root/src/common/mocks/earn.ts
th0rOdinson: Yes, I forgot to add it as a comment. I have validated it with our CTO and we are okay :D

apps/root/src/pages/home/components/token-icon-multichain/index.tsx
th0rOdinson: We used to show network icons (for multi-chain token breakdown). In this case, we would need to show every reward token icon instead of it's network icon

apps/root/src/pages/position-detail/components/summary-container/index.tsx
th0rOdinson: Yes, actually it was implemented as you mention. Since the same method applies to all items, I thought it would be easier to implement it here rather than adding it to each item individually

apps/root/src/common/components/timeline-controls/earn-items/index.tsx
```diff
@@ -35,12 +37,41 @@ import {
...
 import Address from '@common/components/address';
 import { TokenNetworksTooltipTitle } from '@pages/home/components/token-icon-multichain';
 import ComposedTokenIcon from '@common/components/composed-token-icon';
 import { find } from 'lodash';
 import { BalanceToken } from '@hooks/useMergedTokensBalances';
+import { StyledTimelineTitleDate, StyledTimelineTitleEnd } from '../timeline';
+import { DateTime } from 'luxon';
+import { Address as ViemAddress } from 'viem';
+
+const buildDcaTimelineHeader =
```
FiboApe, line 50: 👀  interesting naming 🌝

FiboApe (approved):

PR #920, opened 2024-07-25
Bly 2752

FiboApe:


apps/root/src/pages/token-profile/header/index.tsx
```diff
@@ -0,0 +1,35 @@
+import React from 'react';
+import useNetWorth from '@hooks/useNetWorth';
+import { Token } from 'common-types';
+import TokenIcon from '@common/components/token-icon';
+import { ContainerBox, Typography, colors } from 'ui-library';
+import NetWorthNumber from '@common/components/networth-number';
+
+interface TokenProfileHeaderProps {
+  token?: Token;
```
th0rOdinson, line 9: We can assume token is never undefined right?

apps/root/src/pages/token-profile/token-distribution/index.tsx
```diff
@@ -0,0 +1,290 @@
...
+      background: fill,
+    },
+  })
+);
+
+const DATA_KEY_TO_LABEL: Record<
+  keyof ReturnType<typeof useNetWorth>['assetsTotalValue'],
+  ReturnType<typeof defineMessage>
+> = {
+  dca: defineMessage({ description: 'token-profile.distribution.dca', defaultMessage: 'DCA' }),
```
th0rOdinson, line 115: can't wait to add rows to this record!

th0rOdinson (approved):

apps/root/src/pages/token-profile/header/index.tsx
FiboApe: we could, in case we reuse this I'll prefer to keep it optional for now

PR #923, opened 2024-07-26
feat(token-profile): token based transactions table

th0rOdinson:


apps/root/src/services/transactionService.ts
```diff
@@ -186,38 +223,38 @@ export default class TransactionService extends EventsManager<TransactionService
...
-        transactionsHistory.history = {
-          ...transactionsHistoryResponse,
-          events: [...transactionsHistory.history.events, ...transactionsHistoryResponse.events],
-        };
+      if (!transactionsHistory.history || !isFetchMore) {
+        transactionsHistory.history = { events: [], indexing: {} };
       }
 
+      // Stored events may contain gaps in the timestamp, user may have requested history from different tokens
+      const allEvents = [...transactionsHistory.history?.events, ...transactionsHistoryResponse.events];
```
FiboApe, line 253: history will always be defined here

apps/root/src/pages/home/components/portfolio/index.tsx
```diff
@@ -64,6 +67,26 @@ const StyledNoWallet = styled(ForegroundPaper).attrs({ variant: 'outlined' })`
...
 `;
 
+const StyledTableEnd = styled(TableCell).attrs({ size: 'small' })`
+  ${({ theme: { spacing } }) => `
+    padding: ${spacing(1)} 0px !important;
+  `}
+  height: 1px;
+`;
+
+const StyledLink = styled(Link)`
```
FiboApe, line 77: want to move this to a component? seems like we'll be using it in most tables, like "AnimatedChevron"

apps/root/src/pages/home/components/portfolio/index.tsx
```diff
@@ -64,6 +67,26 @@ const StyledNoWallet = styled(ForegroundPaper).attrs({ variant: 'outlined' })`
   `}
 `;
 
+const StyledTableEnd = styled(TableCell).attrs({ size: 'small' })`
+  ${({ theme: { spacing } }) => `
+    padding: ${spacing(1)} 0px !important;
+  `}
+  height: 1px;
```
FiboApe, line 74: is this actually needed?

apps/root/src/hooks/useTransactionsHistory.tsx
```diff
@@ -82,10 +87,26 @@ function useTransactionsHistory(): {
...
+      if (!!tokens) {
+        // For token history, return only token-related events
+        resolvedFilteredEvents = resolvedEvents.filter(
+          (tx) =>
+            !tokens ||
+            getTransactionInvolvedTokens(tx)
+              .map((token) => getTokenListId({ chainId: token.chainId, tokenAddress: token.address }))
+              .some((tokenId) => tokens.includes(tokenId))
+        );
+      } else if (lastEventTimestamp) {
```
FiboApe, line 101: what about the default case?

th0rOdinson: If lastEventTimestamp doesn't exist, data was never fetched for global events (It should never happen in the use of this hook, thanks to promises initializer). So the empty list should be fine... What do you think?

apps/root/src/pages/home/components/portfolio/index.tsx
th0rOdinson: Yeah! For some reason, we are getting 'rendered more hooks than expected' in virtualized rows. Work around this issue by avoiding the use of hooks

FiboApe (approved):

PR #929, opened 2024-08-01
feat(token-profile): token prices chart

th0rOdinson:
https://github.com/user-attachments/assets/c4808cef-7a96-486b-89cd-58ce3ba49574


apps/root/src/pages/token-profile/historical-prices/index.tsx
```diff
@@ -0,0 +1,356 @@
...
+      // By setting it here, we avoid the flickering of the graph
+      setIsLoadingHistorical(true);
+    }
+
+    setSelectedPeriod(newPeriod);
+  };
+
+  React.useEffect(() => {
+    const handleAllTimeframe = async () => {
+      const firstTimestamp = await providerService.getBlockTimestamp(token.chainId, 1n);
```
th0rOdinson, line 232: We can later explore this endpoint to improve this timestamp lookup:
Screenshot 2024-08-05 at 11 18 39 AM

apps/root/src/services/priceService.ts
```diff
@@ -183,4 +191,59 @@ export default class PriceService {
...
+
+    const prices = [today, daySeconds, weekSeconds, monthSeconds, yearSeconds]
+      .map((seconds) =>
+        this.getUsdHistoricPrice([token], seconds !== today ? (today - seconds).toString() : undefined).then((price) =>
+          Number(formatUnits(price[token.address], 18))
+        )
+      )
+      .map((promise) => timeoutPromise(promise, '10s').catch(console.error));
+
+    const [currentPrice, dayAgoPrice, weekAgoPrice, monthAgoPrice, yearAgoPrice] = await Promise.all(prices);
```
FiboApe, line 238: we could just use Promise.allSettled

apps/root/src/services/priceService.ts
```diff
@@ -183,4 +191,59 @@ export default class PriceService {
...
+
+    const adjustedToken =
+      token.address === PROTOCOL_TOKEN_ADDRESS
+        ? wrappedProtocolToken
+        : { ...token, address: getTokenAddressForPriceFetching(token.chainId, token.address) };
+
+    const tokens = [{ chainId: token.chainId, token: adjustedToken.address }];
+    try {
+      const prices = await timeoutPromise(
+        this.sdkService.sdk.priceService.getChart({
```
FiboApe, line 206: this should be a method of the sdkService itself instead of accesing the sdk property

packages/ui-library/src/theme/variants/button-variants.ts
```diff
@@ -22,6 +22,7 @@ export const buildButtonVariant = (mode: 'light' | 'dark'): Components => ({
       },
       sizeSmall: {
         padding: `${SPACING(1.5)} ${SPACING(3)}`,
+        minWidth: SPACING(8),
```
FiboApe, line 25: hold up cowboy, doesnt this break other stuff?

FiboApe: The small should be as small as it can be

apps/root/src/pages/token-profile/historical-prices/index.tsx
```diff
@@ -0,0 +1,356 @@
...
+      const currentTimestamp = Math.floor(Date.now() / 1000);
+
+      const totalDataPoints = Math.floor((currentTimestamp - firstTimestamp) / (24 * 60 * 60));
```
FiboApe, line 235: so one data point per day?

apps/root/src/pages/token-profile/historical-prices/index.tsx
```diff
@@ -0,0 +1,356 @@
...
+          span: DATA_POINTS,
+        };
+      }
+
+      let span = totalDataPoints;
+
+      let periodDays = 1;
+      while (span > DATA_POINTS) {
+        span = Math.floor(span / 3);
+        periodDays = periodDays * 3;
```
FiboApe, line 250: so if there are a lot of data points we make 1 per 3 days?

apps/root/src/pages/token-profile/historical-prices/index.tsx
```diff
@@ -0,0 +1,356 @@
...
+    }
+  }, [selectedPeriod]);
+
+  const mappedData: DataItem[] = React.useMemo(() => {
+    const prices = graphPrices[selectedPeriod];
+
+    if (!prices) return [];
+
+    const includeYear = [GraphContainerPeriods.year, GraphContainerPeriods.all].includes(selectedPeriod);
+    const format = includeYear ? 'MMM d yyyy' : 'MMM d t';
```
FiboApe, line 281: based on the tickFormatter you are never going to grab the year right?

apps/root/src/pages/token-profile/historical-prices/index.tsx
```diff
@@ -0,0 +1,356 @@
...
+const PeriodMap: Record<Exclude<GraphContainerPeriods, GraphContainerPeriods.all>, TimeString> = {
+  [GraphContainerPeriods.day]: '30m',
+  [GraphContainerPeriods.week]: '3h',
+  [GraphContainerPeriods.month]: '12h',
+  [GraphContainerPeriods.year]: '1w',
+};
+
+const DATA_POINTS = 200;
+
+const TokenHistoricalPrices = ({ token }: TokenHistoricalPricesProps) => {
```
FiboApe, line 211: Seems like there are a looot of thinkgs on this file, could we maybe split it off into several ones so it remains shorter?

apps/root/src/pages/token-profile/historical-prices/index.tsx
```diff
@@ -0,0 +1,356 @@
...
+        span,
+      };
+    };
+
+    const fetchGraphPrices = async () => {
+      const priceResult = await priceService.getPricesForTokenGraph({
+        token,
+        ...(selectedPeriod === GraphContainerPeriods.all
+          ? await handleAllTimeframe()
+          : { period: PeriodMap[selectedPeriod], span: DATA_POINTS }),
```
FiboApe, line 264: Should we actually define how many data points we want based on the selected period?

FiboApe: And btw, we could extend longer periods with the data of smaller periods if we have already loaded them

apps/root/src/pages/token-profile/historical-prices/index.tsx
```diff
@@ -0,0 +1,356 @@
...
+      });
+
+      setGraphPrices((prev) => ({ ...prev, [selectedPeriod]: priceResult }));
```
FiboApe, line 267: see, here we dont need to have results per period, we could just have results and extend those when a period gets selected

apps/root/src/pages/token-profile/historical-prices/index.tsx
```diff
@@ -0,0 +1,356 @@
...
+        defaultEnabledPeriods={[
+          GraphContainerPeriods.day,
+          GraphContainerPeriods.week,
+          GraphContainerPeriods.month,
+          GraphContainerPeriods.year,
+          GraphContainerPeriods.all,
+        ]}
+        defaultPeriod={GraphContainerPeriods.month}
+        minHeight={270}
+        updatePeriodCb={handlePeriodUpdate}
```
FiboApe, line 306: lets use the full name callback plz

apps/root/src/pages/token-profile/historical-prices/index.tsx
th0rOdinson: Here, when 'All' period is selected, I calcualte how many days there are between the first block and today. Later, in the while loop, I reduce this number into something adjusted to the graph

apps/root/src/pages/token-profile/historical-prices/index.tsx
th0rOdinson: This `while` ensures the number is between 60 and 200 aproximately. So if there are a LOT of data points, we reduce them until we are below 200
For instance, in the case of WETH, we have 3295 days (6,5 years). So we request a `period` of `27d` with a `span` of `122`.

apps/root/src/pages/token-profile/historical-prices/index.tsx
th0rOdinson: Yes, but the design specifies x axis tick should be hidden (I agree, don't like it when its rendered)
Screenshot 2024-08-07 at 12 23 10 PM
Date data is only displayed in the tooltip

apps/root/src/pages/token-profile/historical-prices/index.tsx
th0rOdinson: Sure, I'll work on that now

apps/root/src/services/priceService.ts
th0rOdinson: Absolutely! This wasn't available due to @types/node version, but now we can use it :D

packages/ui-library/src/theme/variants/button-variants.ts
th0rOdinson: You're right! This was just migrated from feat/earn, but 0 as min-width works just fine

apps/root/src/pages/token-profile/historical-prices/index.tsx
th0rOdinson: I spent some time trying to find the 'universal query' for this graph, but didn't get to any result in that logic. I have refined a bit how I construct the periods based on a span of `200`, keeping this examples in mind:
- For `1d` as selected period, I'm showing data within 24 hours (period: 10m, span: 200 points => 2000 minutes => 30 hours)
- For `1y` as selected period, I'm showing data in a whole year (period: 2d, span: 200 points => 400 days => 1,09 years) 

I think SOME points can be recycled, but the graph would need to refetch some specific data in order to render a more precise chart

apps/root/src/pages/token-profile/historical-prices/index.tsx
FiboApe: Mmmm I think we should display it, you can fiddle around with the tickMargin I think or some other prop, to show less of thems, but it seems important to have the dates applied

apps/root/src/pages/token-profile/historical-prices/index.tsx
FiboApe: It makes a bit more sense, I think we should still render more points when its a more longer period, having the daily selected as default would allow us to keep the data more consistent

apps/root/src/pages/token-profile/historical-prices/index.tsx
FiboApe: Do you know if the query takes longer if we have a smaller period?

th0rOdinson: Couldn't find a strict relation between period and response time

apps/root/src/pages/token-profile/historical-prices/index.tsx
th0rOdinson: Will do!

apps/root/package.json
```diff
@@ -35,7 +35,7 @@
     "@types/luxon": "^1.27.0",
     "@types/md5": "^2.3.2",
     "@types/mixpanel-browser": "2.38.1",
-    "@types/node": "^15.6.1",
+    "@types/node": "^22.1.0",
```
FiboApe, line 38: Match this to the UI library one please

    "@types/node": "^20.5.2",

FiboApe (approved):

PR #932, opened 2024-08-07
feat(aggregator): handle quotes asynchronicously

FiboApe:


th0rOdinson (commented):
FANTASTIC job! I've found something that we could improve in this PR:
1. When you just enter the site, and select from: protocol to: any other, after quotes load, there are 2 seconds where we are showing no loading state, but no quote is selected
2. Same happens when a quote source was selected, it takes some time to override it, probably because of simulation time

th0rOdinson (approved):

PR #933, opened 2024-08-07
feat(earn): portfolio table

th0rOdinson:


apps/root/src/state/strategies-filters/reducer.ts
```diff
@@ -47,15 +47,23 @@ const initialFilters = {
...
-  orderBy: {
-    column: StrategyColumnKeys.TVL,
-    order: 'desc' as ColumnOrder,
-  },
 };
 
 export const initialState: StrategiesFiltersState = {
-  [StrategiesTableVariants.ALL_STRATEGIES]: { ...initialFilters },
-  [StrategiesTableVariants.USER_STRATEGIES]: { ...initialFilters },
+  [StrategiesTableVariants.ALL_STRATEGIES]: {
```
FiboApe, line 53: should we actually have two redux stores for this?

apps/root/src/pages/earn/portfolio/frame/index.tsx
```diff
@@ -7,16 +7,46 @@ import { EARN_PORTFOLIO } from '@constants/routes';
...
   const [selectedWalletOption, setSelectedWalletOption] = React.useState<WalletOptionValues>(ALL_WALLETS);
+  const { userStrategies, hasFetchedUserStrategies } = useEarnPositions();
 
   const dispatch = useAppDispatch();
 
   React.useEffect(() => {
     dispatch(changeRoute(EARN_PORTFOLIO.key));
   }, []);
 
+  const groupedPositionsByStrategy = React.useMemo(() => {
```
FiboApe, line 29: I'm kinda on the fence on this, since it's going to re-render EVERYTHING when any strategy changes. We can have a wrapper component for the StrategiesTable that just uses the hook internally so only that component re-renders

apps/root/src/pages/earn/home/frame/index.tsx
```diff
@@ -10,14 +10,15 @@ import useEarnService from '@hooks/earn/useEarnService';
...
+import { strategyColumnConfigs } from '@pages/earn/components/strategies-table/components/columns';
 
 const variant = StrategiesTableVariants.ALL_STRATEGIES;
 
 const EarnFrame = () => {
   const dispatch = useAppDispatch();
   const earnService = useEarnService();
   const { strategies, hasFetchedAllStrategies } = useAllStrategies();
-  const filteredStrategies = useFilteredStrategies({ variant, strategies });
+  const filteredStrategies = useFilteredStrategies({ variant, strategies, columns: strategyColumnConfigs });
```
FiboApe, line 21: Same comment could be applied here if we are not using filtered strategies for anything el se here

apps/root/src/pages/earn/components/strategies-table/index.tsx
```diff
@@ -1,5 +1,5 @@
 import React from 'react';
-import { Strategy } from 'common-types';
+import { EarnPosition, Strategy } from 'common-types';
```
FiboApe, line 2: I like all the generalization here, tho I'm a bit afraid it's gonna bite us in the ass later when we want to extend it a bit more.

For now lets keep is as you made it

apps/root/src/pages/earn/components/strategies-table/components/columns/index.tsx
```diff
@@ -20,25 +32,93 @@ export enum StrategyColumnKeys {
...
+  );
+};
+
+export interface StrategyColumnConfig<T extends StrategiesTableVariants> {
   key: StrategyColumnKeys;
   label: React.ReactNode;
-  renderCell: (data: TableStrategy) => React.ReactNode | string;
-  getOrderValue?: (data: TableStrategy) => string | number;
+  renderCell: (data: TableStrategy<T>) => React.ReactNode | string;
+  customCell?: boolean;
```
FiboApe, line 116: custom would just mean its a reactNode that is being sent or why do we have this boolean?

apps/root/src/hooks/earn/useFilteredStrategies.ts
```diff
@@ -5,23 +5,27 @@ import { searchByStrategyData } from '@common/utils/earn/search';
 import { StrategiesTableVariants } from '@state/strategies-filters/reducer';
 import { getComparator } from '@common/utils/earn/parsing';
 import { EarnPosition, Strategy } from 'common-types';
+import { StrategyColumnConfig } from '@pages/earn/components/strategies-table/components/columns';
 
-const isEarnPosition = (obj: Strategy | EarnPosition): obj is EarnPosition => 'strategy' in obj;
+const isEarnPosition = (obj: Strategy | EarnPosition[], variant: StrategiesTableVariants): obj is EarnPosition[] =>
```
FiboApe, line 10: this doesn't actually use the obj right? just the variant

apps/root/src/pages/earn/components/strategies-table/components/columns/index.tsx
th0rOdinson: It was for adding a custom padding for bigger cells, but now it's gone :)

apps/root/src/hooks/earn/useFilteredStrategies.ts
th0rOdinson: It's used for the type assertion, so I can get rid of the union type and differenciate one case from the other

FiboApe (approved):

PR #936, opened 2024-08-12
Improve wallet clients handling

FiboApe:


apps/root/src/hooks/useClaimableCampaigns.ts
```diff
@@ -25,79 +24,66 @@ function useClaimableCampaigns(): [Campaigns | undefined, boolean, string?] {
...
 
   React.useEffect(() => {
     async function callPromise() {
-      if (!account) {
-        return;
-      }
-
       try {
-        const promiseResult = await campaignService.getCampaigns(account);
+        const promiseResult = await Promise.all(wallets.map(({ address }) => campaignService.getCampaigns(address)));
```
th0rOdinson, line 36: I know this file is not used today, but we can change this to allSettled

th0rOdinson (commented):
Truly amazing!

th0rOdinson (approved):

apps/root/src/common/components/wallet-selector/index.tsx
```diff
@@ -392,15 +405,9 @@ const WalletSelector = ({ options, size = 'small' }: WalletSelectorProps) => {
...
-      <OptionsMenu
-        options={[connectWalletOption]}
-        mainDisplay={intl.formatMessage(
-          defineMessage({
-            defaultMessage: 'Connect your wallet',
-            description: 'connectWallet',
-          })
-        )}
-      />
+      <Button onClick={onConnectWallet} variant="text" size="small">
```
th0rOdinson, line 408: I think this would look better with padding in 4px, to match the styles of the connected state

th0rOdinson (commented):
Found a new bug, doesn't happen every time, but still, steps to reproduce:
- Enter Balmy being already signed out
- Connect and sign message (im ussing rabby)
- Disconnect, and then connect again.
Rabby only asks for connect, not for a new signature, so user never logs in again

apps/root/src/common/components/wallet-selector/index.tsx
FiboApe: Yeah this is just a test for now

PR #938, opened 2024-08-16
feat(earn): home widget

th0rOdinson:
### Desktop
Screenshot 2024-08-16 at 11 30 04 AM

### Mobile
Screenshot 2024-08-16 at 11 23 05 AM

### DIFF
--- a/apps/root/src/pages/home/components/earn-positions-dashboard/index.tsx
+++ b/apps/root/src/pages/home/components/earn-positions-dashboard/index.tsx
@@ -1,0 +1,102 @@
+import React from 'react';
+import { ALL_WALLETS, WalletOptionValues } from '@common/components/wallet-selector';
+import useNetWorth from '@hooks/useNetWorth';
+import WidgetFrame from '../widget-frame';
+import { FormattedMessage } from 'react-intl';
+import { colors, ContainerBox, Grid, MoneyAddIcon, Typography } from 'ui-library';
+import useEarnPositions from '@hooks/earn/useEarnPositions';
+import { useShowBalances } from '@state/config/hooks';
+import ExpectedReturns from '@pages/strategy-guardian-detail/investment-data/components/expected-returns';
+import FinancialOverview from '@pages/strategy-guardian-detail/investment-data/components/financial-overview';
+
+interface EarnPositionsDashboardProps {
+  selectedWalletOption: WalletOptionValues;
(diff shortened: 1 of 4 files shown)


apps/root/src/pages/home/components/earn-positions-dashboard/index.tsx
```diff
@@ -0,0 +1,102 @@
...
+                />
+              </Grid>
+              <Grid item xs={12} md="auto" display="flex" flexDirection="column" gap={3}>
+                <Typography variant="bodyBold" color={({ palette }) => colors[palette.mode].typography.typo1}>
+                  <FormattedMessage
+                    defaultMessage="Expected returns"
+                    description="home.earn.dashboard.title.expected-returns"
+                  />
+                </Typography>
+                <ExpectedReturns userPositions={filteredPositions} size="small" isLoading={!hasFetchedUserStrategies} />
```
th0rOdinson, line 85: In `ExpectedReturns` and `FinancialOverview`, size=small gives a gap of `6`. For this case, we should use `8`, but it would look like this:
Screenshot 2024-08-16 at 11 32 44 AM
This occurs because figma's max-width is bigger than our App, this is a known difference we already discussed
I think this is 100% acceptable imho, just wanted to give visibility :)

FiboApe: God I love how we reuse components

apps/root/src/pages/home/frame/index.tsx
```diff
@@ -73,11 +74,8 @@ const HomeFrame = () => {
                   </StyledContent>
                 </StyledContainer>
               </Grid>
-              <Grid item xs={12} display="flex">
```
FiboApe, line 76: Is there a reason why we removed this from here and sit it on each component? Would prefer to have the defining structure in this file directly

th0rOdinson: Moved it to the component itself, because it still takes place in the UI when user has no positions, as an empty grid item. I know structure is more clear, but we can handle conditional rendering in the component

FiboApe: MM i get that but I would rather keep all structure related to where we display the grids just for clarity

FiboApe (approved):

PR #939, opened 2024-08-19
Earn deposit!

FiboApe:


apps/root/src/pages/strategy-guardian-detail/strategy-management/deposit/tx-manager/index.tsx
```diff
@@ -0,0 +1,68 @@
...
+        handleClose={handleBackTransactionSteps}
+        transactions={transactionSteps}
+        onAction={transactionOnAction.onAction}
+        onActionConfirmed={transactionOnAction.onActionConfirmed}
+        applicationIdentifier={TransactionApplicationIdentifier.EARN_DEPOSIT}
+        setShouldShowFirstStep={() => {}}
+        setHeight={setHeight}
+        recapDataProps={recapDataProps}
+      />
+      <EarnDepositTransactionConfirmation
```
th0rOdinson, line 56: I think this cases are when we should call it deposit, and then differenciate with create/increase

apps/root/src/pages/strategy-guardian-detail/strategy-management/deposit/tx-steps/index.tsx
```diff
@@ -0,0 +1,16 @@
+import React from 'react';
+import TransactionSteps from '@common/components/transaction-steps';
+import { useEarnManagementState } from '@state/earn-management/hooks';
+
+const EarnTransactionSteps = ({ recapDataProps, ...rest }: Parameters<typeof TransactionSteps>[0]) => {
+  const { assetAmount } = useEarnManagementState();
+
+  const earnRecapDataProps = React.useMemo(
+    () => (recapDataProps ? { ...recapDataProps, assetAmount } : undefined),
+    [assetAmount]
```
th0rOdinson, line 10: Same here, don't we why we subscribe here to redux and memoize it, if we can consume directly in `EarnDepositRecapData`

apps/root/src/services/earnService.ts
```diff
@@ -389,4 +400,275 @@ export class EarnService extends EventsManager<EarnServiceData> {
...
+      default:
+        userStrategies = [...this.userStrategies];
+        break;
+    }
+
+    this.userStrategies = userStrategies;
+  }
+
+  handleTransaction(transaction: TransactionDetails) {
+    if (transaction.type !== TransactionTypes.earnDeposit && transaction.type !== TransactionTypes.earnIncrease) {
```
th0rOdinson, line 539: Oh yeah, this logic is what we need in DCA!

apps/root/src/services/earnService.ts
```diff
@@ -389,4 +400,275 @@ export class EarnService extends EventsManager<EarnServiceData> {
...
+        break;
+      default:
+        userStrategies = [...this.userStrategies];
+        break;
+    }
+
+    this.userStrategies = userStrategies;
+  }
+
+  handleTransaction(transaction: TransactionDetails) {
```
th0rOdinson, line 538: In this method we should also care for the fake BigInt's from localStorage

apps/root/src/pages/strategy-guardian-detail/strategy-management/deposit/hooks/useEarnDepositActions.tsx
```diff
@@ -0,0 +1,563 @@
...
+            defineMessage({
+              description: 'approveTokenExplanation',
+              defaultMessage:
+                'By enabling Universal Approval, you will be able to use Uniswap, Balmy, swap aggregators and more protocols without having to authorize each one of them',
+            })
+          ),
+          extraData: {
+            token: asset,
+            amount: assetAmount,
+            swapper: intl.formatMessage(
```
th0rOdinson, line 449: Don't see the use of swapper prop here

apps/root/src/pages/strategy-guardian-detail/strategy-management/deposit/hooks/useEarnDepositActions.tsx
```diff
@@ -0,0 +1,563 @@
...
+      currentTransaction,
+      transactionOnAction,
+      handleBackTransactionSteps,
+      onDeposit,
+      handleMultiSteps,
+    ]
+  );
+};
+
+export default useEarnDepositActions;
```
th0rOdinson, line 563: Really liked this logic abstraction! Can't wait to implement it in DCA & Agg

apps/root/src/services/earnService.spec.ts
```diff
@@ -0,0 +1,627 @@
...
+          timestamp: now - 10,
+          tx: {
+            hash: '0xhash',
+            timestamp: now - 10,
+          },
+        },
+      ],
+});
+
+describe('Position Service', () => {
```
th0rOdinson, line 201: Earn service

th0rOdinson (commented):
Heroic PR, congratulations!

packages/common-types/src/transactions.ts
```diff
@@ -395,4 +423,5 @@ export enum TransactionApplicationIdentifier {
   TRANSFER = 'TRANSFER',
   SWAP = 'SWAP',
   DCA = 'DCA',
+  EARN_DEPOSIT = 'EARN_DEPOSIT',
```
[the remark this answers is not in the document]
FiboApe: Yeah we should actually have both here since we use them for the satisfaction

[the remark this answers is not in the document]
FiboApe: As for the naming, the method on the SC will always be called deposit for a first time deposit so kinda following that pattern, since the function on the earnService is also going to be called deposit

apps/root/src/pages/strategy-guardian-detail/strategy-management/deposit/recap-data/index.tsx
```diff
@@ -0,0 +1,130 @@
...
+)``;
+
+interface EarnDepositRecapDataProps {
+  strategy?: DisplayStrategy;
+  assetAmount?: string;
+}
+
+const EarnDepositRecapData = ({ strategy, assetAmount: assetAmountInUnits }: EarnDepositRecapDataProps) => {
+  const intl = useIntl();
+  const token = useToken({ tokenAddress: strategy?.asset.address, chainId: strategy?.network.chainId });
```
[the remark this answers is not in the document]
FiboApe: I think we already talked about this :)

apps/root/src/pages/strategy-guardian-detail/strategy-management/deposit/form/index.tsx
```diff
@@ -0,0 +1,32 @@
...
+  const activeWallet = useActiveWallet();
+  const { balance } = useTokenBalance({
+    token: strategy?.asset || null,
+    walletAddress: activeWallet?.address,
+    shouldAutoFetch: true,
+  });
+
+  return (
+    <ContainerBox gap={3} flexDirection="column">
+      <WalletSelect />
```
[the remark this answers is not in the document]
FiboApe: Yeah this is gonna get replaced by the one you created

apps/root/src/common/components/wallet-select/index.tsx
```diff
@@ -0,0 +1,119 @@
...
+const WalletItem = ({ item: wallet }: { item: OptionWithKey }) => {
+  const mode = useThemeMode();
+  const intl = useIntl();
+
+  return (
+    <ContainerBox alignItems="center" justifyContent="space-between" key={wallet.key} flex={1} gap={3}>
+      <ContainerBox alignItems="center" flex={1} gap={3}>
+        <WalletIcon />
+        <Typography variant="bodySemibold" color={colors[mode].typography.typo2}>
+          <Address address={wallet.address} trimAddress />
```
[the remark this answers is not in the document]
FiboApe: Same as before

apps/root/src/common/utils/transaction-history/index.ts
```diff
@@ -89,6 +99,10 @@ export const getTransactionTokenFlow = (tx: TransactionEvent, wallets: string[])
     case TransactionEventTypes.DCA_CREATED:
       return TransactionEventIncomingTypes.INCOMING;
       break;
+    case TransactionEventTypes.EARN_DEPOSITED:
```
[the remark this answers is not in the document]
FiboApe: Mmm i would prefer deposit since its the sc function

apps/root/src/pages/strategy-guardian-detail/strategy-management/deposit/tx-steps/index.tsx
FiboApe: I'm sorry i didnt understand the comment

apps/root/src/services/earnService.ts
FiboApe: Made it better still :D

apps/root/src/pages/strategy-guardian-detail/strategy-management/deposit/hooks/useEarnDepositActions.tsx
FiboApe: its needed for the type of the approve

packages/common-types/src/transactions.ts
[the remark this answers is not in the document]
th0rOdinson: Okay, but i'm seeing that we are using `deposit` for both terms: deposit (create + increase) and just create. I think we should not use deposit term when we want to differentiate create and increase (in transaction history for example), just to prevent asuming that the 'deposit' name is englobing all

packages/common-types/src/transactions.ts
```diff
@@ -53,22 +49,103 @@ export enum TransactionTypes {
...
+  TransactionTypes.resetPosition,
+  TransactionTypes.eulerClaimTerminateMany,
+  TransactionTypes.eulerClaimPermitMany,
+  TransactionTypes.eulerClaimApproveMigrator,
+  TransactionTypes.eulerClaimClaimFromMigrator,
+] as const;
+
+export type DcaTransactionTypes = (typeof DcaTransactionArrayTypes)[number];
+
+export function isDcaType(
```
th0rOdinson, line 95: Ohhhhhhhh

apps/root/src/pages/strategy-guardian-detail/strategy-management/deposit/form/index.tsx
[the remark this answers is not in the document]
th0rOdinson: Okay, but take into account that the current component only filters by selected network. This is the first case where we know for sure the network & token, so another filter shall be added (it's now available in feat/earn)

apps/root/src/pages/strategy-guardian-detail/strategy-management/deposit/tx-steps/index.tsx
th0rOdinson: This is resolved now, it was in relation with accessing redux directly in the recap data component, but I agree its more useful now via props

th0rOdinson (approved):

PR #941, opened 2024-08-21
feat(history): add indexing notices when indexes are not complete

FiboApe:


apps/root/src/common/utils/transaction-history/parsing.tsx
```diff
@@ -348,6 +353,7 @@ const parseDcaTransferApiEvent: ParseFunction<DCATransferApiEvent, DCATransferEv
 }) => {
   const parsedEvent: DCATransferEvent = {
     type: TransactionEventTypes.DCA_TRANSFER,
+    unit: IndexerUnits.DCA,
```
th0rOdinson, line 356: As a suggestion, another approach would be having a Record<TransactionEventTypes, IndexerUnits> in order to get this data for each event? Instead of adding them to each event, I find it simpler
We have a DCA_TYPE_EVENTS list that can be useful too

apps/root/src/services/transactionService.ts
```diff
@@ -214,22 +226,22 @@ export default class TransactionService extends EventsManager<TransactionService
...
-      >((acc, [address, chainsData]) => {
-        if (!('error' in chainsData)) {
-          Object.entries(chainsData).forEach(([chainId, chainData]) => {
-            // eslint-disable-next-line no-param-reassign
-            acc[address as Address] = { ...acc[address as Address], [Number(chainId)]: chainData };
-          });
-        }
-        return acc;
-      }, {});
-
```
th0rOdinson, line 232: Not sure why we moved this into both if-else statements

FiboApe: Moved it out :)

th0rOdinson (approved):

PR #942, opened 2024-08-21
fix(token-profile): feedback fixes

th0rOdinson:
### Fixes addressed in this PR:
- [x] Try to adjust gap between last two X tickers
- [x] Add margin-left to X axis tickers
- [x] Add amountInUnits to portfolio table
- [x] Decimals in Net Worth should have less opacity
- [x] Rename distribution to be Spot and Recurring Investments => Created **`BLY-2952`** for improvement
- [x] Add % to distribution numbers
- [x] The Home menu item is highlighted when token profile page is open
- [x] Remove 100% from portfolio table
- [x] Change Top NetWorth number to be the token price
- [x] Explorer links should take to the address, not the token page => Dismissed
- [x] Set Default graph period to 1 week
- [x] Graph paper color
- [x] DCA is loading all net worth, not filtered by token
- [x] Make full portfolio row clickable and hoverable

### Pending issues:
- [ ] Debug on failed prices from history => Created **`BLY-2955`**
- [ ] Debug old timestamps (LINK from +1 year ago) => Created **`BLY-2953`**

Created and added to **`BLY-2954`**:
- [ ] Add event date to GraphTooltip
- [ ] Improve skeleton loading
- [ ] Below 'All Wallets' we have a border-bottom that shouldn't be there
- [ ] Rows in AllWallets table have a border-radius and they shouldn't
- [ ] Remove border-bottom from last row
- [ ] Add hover to each explorer Chip, like we have in Activity (home)
- [ ] Try to make Y trickers equally distant

### DIFF
--- a/apps/root/src/common/components/net-worth/index.tsx
+++ b/apps/root/src/common/components/net-worth/index.tsx
@@ -65,7 +65,6 @@ const NetWorth = ({ walletSelector, chainId }: NetWorthProps) => {
         withAnimation
         value={totalAssetValue}
         variant="h4Bold"
-        addDolarSign
         size="large"
       />
     </StyledNetWorthContainer>
--- a/apps/root/src/constants/routes.tsx
+++ b/apps/root/src/constants/routes.tsx
@@ -30,5 +30,9 @@ export const TRANSFER_ROUTE = {
   icon: <TransferIcon />,
 };
 
-export const TOKEN_PROFILE_ROUTE = 'token';
-export const HISTORY_ROUTE = 'history';
+export const TOKEN_PROFILE_ROUTE = {
+  key: 'token',
+};
+export const HISTORY_ROUTE = {
+  key: 'history',
+};
--- a/apps/root/src/constants/routes.tsx
+++ b/apps/root/src/constants/routes.tsx
@@ -29,3 +29,6 @@ export const TRANSFER_ROUTE = {
   key: 'transfer',
   icon: <TransferIcon />,
 };
+
+export const TOKEN_PROFILE_ROUTE = 'token';
+export const HISTORY_ROUTE = 'history';
--- a/apps/root/src/hooks/useNetWorth.ts
+++ b/apps/root/src/hooks/useNetWorth.ts
@@ -83,7 +83,14 @@ const useNetWorth = ({ walletSelector, chainId, tokens }: NetWorthProps) => {
   }
 
   let dcaAssetsTotalValue = 0;
-  let filteredPositions = currentPositions;
+  let filteredPositions = currentPositions.filter(
+    (position) =>
+      !(tokens && tokens.length) ||
+      tokens.some(
+        (filterToken) =>
+          getIsSameOrTokenEquivalent(filterToken, position.from) || getIsSameOrTokenEquivalent(filterToken, position.to)
+      )
+  );
   if (walletSelector !== ALL_WALLETS) {
     if (chainId && walletAddressToEvaluate) {
       filteredPositions = currentPositions.filter(
--- a/apps/root/src/pages/home/components/portfolio/index.tsx
+++ b/apps/root/src/pages/home/components/portfolio/index.tsx
@@ -57,6 +57,7 @@ import TokenIconMultichain from '../token-icon-multichain';
 import useAccountService from '@hooks/useAccountService';
 import { Link } from 'react-router-dom';
 import usePushToHistory from '@hooks/usePushToHistory';
+import { TOKEN_PROFILE_ROUTE } from '@constants/routes';
 
 const StyledNoWallet = styled(ForegroundPaper).attrs({ variant: 'outlined' })`
   ${({ theme: { spacing } }) => `
@@ -177,7 +178,7 @@ const PortfolioNotConnected = () => {
 const PortfolioBodyItem: ItemContent<BalanceItem, Context> = (
   index: number,
   { totalBalanceInUnits, tokens, isLoadingPrice, price, totalBalanceUsd, relativeBalance }: BalanceItem,
-  { intl, showBalances, hoveredRow }
+  { intl, showBalances }
 ) => {
   const firstAddedToken = tokens[0].token;
   return (
@@ -250,7 +251,7 @@ const PortfolioBodyItem: ItemContent<BalanceItem, Context> = (
         )}
         <StyledTableEnd>
           <StyledLink to={`/token/${firstAddedToken.chainId}-${firstAddedToken.address}`}>
-            <AnimatedChevronRightIcon $hovered={hoveredRow === index} />
+            <AnimatedChevronRightIcon />
           </StyledLink>
         </StyledTableEnd>
       </Hidden>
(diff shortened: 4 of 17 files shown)


apps/root/src/hooks/useNetWorth.ts
```diff
@@ -83,7 +83,14 @@ const useNetWorth = ({ walletSelector, chainId, tokens }: NetWorthProps) => {
   }
 
   let dcaAssetsTotalValue = 0;
-  let filteredPositions = currentPositions;
```
FiboApe, line 86: yeah that explains it haha

packages/ui-library/src/components/virtualized-table/index.tsx
```diff
@@ -59,7 +59,13 @@ const StyledBodySmallLabelTypography = styled(Typography).attrs(
   })
 )``;
 
-interface BaseContext {}
+interface BaseContext {
```
FiboApe, line 62: Could we change the name to something more descriptive? Like VirtualizedTableContext

packages/ui-library/src/components/virtualized-table/index.tsx
```diff
@@ -86,7 +92,15 @@ function buildVirtuosoTableComponents<D, C extends BaseContext>(): TableComponen
     }),
     Table: (props) => <Table sx={{ padding: 0 }} {...props} />,
     TableHead,
-    TableRow: ({ item: _item, ...props }) => <TableRow {...props} />,
+    TableRow: ({ item: _item, context, ...props }) => (
+      <TableRow
+        {...props}
+        onMouseEnter={() => context?.rowContext?.setHovered && context.rowContext.setHovered(props['data-index'])}
```
FiboApe, line 98: What do we need this for? and where are we sending data-index?

packages/ui-library/src/icons/animated-chevron-right.tsx
```diff
@@ -8,16 +8,11 @@ const AnimatedChevronRight = styled((props: SvgIconProps) => (
...
+  ${({ $hovered, theme: { spacing } }) => `
   transition: transform 0.15s ease;
-  ${
-    $controlled
-      ? `transform: translateX(${$hovered ? spacing(1) : 0});`
-      : `&:hover {
-          transform: translateX(${spacing(1)});
-        }`
-  }`}
+  transform: translateX(${$hovered ? spacing(1) : 0});
```
FiboApe, line 14: but this should only be applied on hover right?

apps/root/src/pages/home/components/portfolio/index.tsx
```diff
@@ -399,12 +403,34 @@ const Portfolio = ({ selectedWalletOption }: PortfolioProps) => {
...
+  );
+
+  const tableContext = React.useMemo<Context>(
+    () => ({
+      intl,
+      showBalances,
+      hoveredRow,
+      rowContext: {
+        onClick: onRowClick,
+        setHovered: setHoveredRow,
```
FiboApe, line 421: I think this should be handled directly by the component, I get the we mostly do it for the chevron animation, but we could do that by using styles and pointing to this component when the main row gets hovered on the virtualized table

apps/root/src/pages/token-profile/frame/index.tsx
```diff
@@ -36,11 +39,12 @@ const TokenProfileFrame = () => {
   });
 
   React.useEffect(() => {
+    dispatch(changeRoute('token'));
```
FiboApe, line 42: can we use the ROUTE key for this?

apps/root/src/pages/token-profile/token-distribution/index.tsx
```diff
@@ -91,29 +89,12 @@ const StyledBullet = styled.div<{ fill: string }>`
...
-  })
-);
-
 const DATA_KEY_TO_LABEL: Record<
   keyof ReturnType<typeof useNetWorth>['assetsTotalValue'],
   ReturnType<typeof defineMessage>
 > = {
-  dca: defineMessage({ description: 'token-profile.distribution.dca', defaultMessage: 'DCA' }),
-  wallet: defineMessage({ description: 'token-profile.distribution.wallet', defaultMessage: 'Wallet' }),
+  dca: defineMessage({ description: 'token-profile.distribution.dca', defaultMessage: 'Recurring Investments' }),
```
FiboApe, line 96: We generally dont need to change this since it can be later updated through localazy, its fine for now just future reference

apps/root/src/common/components/net-worth/index.tsx
```diff
@@ -65,7 +65,6 @@ const NetWorth = ({ walletSelector, chainId }: NetWorthProps) => {
         withAnimation
         value={totalAssetValue}
         variant="h4Bold"
-        addDolarSign
```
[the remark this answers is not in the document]
th0rOdinson: isFiat is true by default, I find it as the most common use case for this number

packages/ui-library/src/icons/animated-chevron-right.tsx
th0rOdinson: Isn't it like that? TransateX: 0 should not change the position of the element

packages/ui-library/src/components/virtualized-table/index.tsx
th0rOdinson: As the hover and click events should be applied to the whole row, I didn't find any way to do this in the consumer component (We only send the children of TableRow for the table body). The only way I found was to apply this props when building the table

apps/root/src/pages/home/components/portfolio/index.tsx
th0rOdinson: Answered above!

apps/root/src/pages/home/components/widget-frame/index.tsx
```diff
@@ -109,7 +109,6 @@ const WidgetFrame = ({
           </Typography>
           <NetWorthNumber
             value={assetValue}
-            addDolarSign
```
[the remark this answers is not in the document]
th0rOdinson: Same as before, isFiat is true by default  :D

apps/root/src/pages/token-profile/frame/index.tsx
th0rOdinson: Sure, also this can be improved in 'history' page

apps/root/src/pages/token-profile/token-distribution/index.tsx
th0rOdinson: I agree with you, but we received feedback based on this, and I was able to find out that the component wasn't prepared for the real wording we are going to be using in production

apps/root/src/constants/routes.tsx
```diff
@@ -29,3 +29,6 @@ export const TRANSFER_ROUTE = {
   key: 'transfer',
   icon: <TransferIcon />,
 };
+
+export const TOKEN_PROFILE_ROUTE = 'token';
```
FiboApe, line 33: lets make this an object even tho we dont show it on the navigation to keep consistency

FiboApe (approved):

packages/ui-library/src/components/virtualized-table/index.tsx
```diff
@@ -92,13 +99,10 @@ function buildVirtuosoTableComponents<D, C extends VirtualizedTableContext>(): T
     }),
     Table: (props) => <Table sx={{ padding: 0 }} {...props} />,
     TableHead,
-    TableRow: ({ item: _item, context, ...props }) => (
-      <TableRow
+    TableRow: ({ item, context, ...props }) => (
+      <StyledRow
+        onClick={context?.onRowClick ? () => context.onRowClick?.(props['data-index']) : undefined}
```
FiboApe, line 104: onRowClick exists at this point :)

PR #943, opened 2024-08-21
feat(beamer): add beamer whats new and its badge icon

FiboApe:


packages/ui-library/src/components/navigation/index.tsx
```diff
@@ -335,7 +335,12 @@ const Navigation = ({
             </IconButton>
             <AppBarRightContainer alignItems="center" justifyContent="flex-end" flex={1} gap={2}>
               {extraHeaderTools}
-              <OptionsMenu options={helpOptions} mainDisplay={<HelpIcon />} />
+              <OptionsMenu
+                options={helpOptions}
+                customClassname="beamer-whats-new"
```
th0rOdinson, line 340: Is it necessary to have this class here? It's already applied to the option item

FiboApe: We need this so the badge also shows up on the help icon

th0rOdinson (approved):

PR #946, opened 2024-08-22
feat(home): News Banner

th0rOdinson:
Screenshot 2024-08-22 at 3 56 28 PM


FiboApe:
Check with javi but If I remember correctly it should go on the bottom of the navigation

apps/root/src/pages/home/frame/index.tsx
```diff
@@ -39,24 +41,50 @@ const HomeFrame = () => {
...
   const trackEvent = useTrackEvent();
+  const pushToHistory = usePushToHistory();
+  const intl = useIntl();
 
   React.useEffect(() => {
     dispatch(changeRoute(DASHBOARD_ROUTE.key));
     trackEvent('Home - Visit Dashboard Page');
   }, []);
 
+  const newsBannerProps = {
```
FiboApe, line 52: We should send this all into its own component, so we only need to change that once we change the banner

FiboApe (approved):

PR #949, opened 2024-08-23
feat(earn): add guardian section

FiboApe:


apps/root/src/pages/strategy-guardian-detail/vault-data/components/data-guardian/index.tsx
```diff
@@ -0,0 +1,164 @@
...
+    display: flex;
+  `}
+`;
+
+const StyledGuardianPill = styled(ContainerBox).attrs({ gap: 2, alignItems: 'center' })`
+  ${({ theme }) => `
+    padding: ${theme.spacing(2)} ${theme.spacing(4)} ${theme.spacing(2)} ${theme.spacing(2)};
+    background-color: ${colors[theme.palette.mode].background.secondary};
+    border: 1.5px solid ${colors[theme.palette.mode].border.border1};
+    border-radius: ${theme.spacing(15)}
```
th0rOdinson, line 54: Should add `;` at the end

apps/root/src/pages/strategy-guardian-detail/vault-data/components/data-guardian/index.tsx
```diff
@@ -0,0 +1,164 @@
...
+  Skeleton,
+  ArrowDropDownIcon,
+  ArrowDropUpIcon,
+  Collapse,
+  ForegroundPaper,
+  Link,
+  TwitterIcon,
+  DiscordIcon,
+} from 'ui-library';
+import { SPACING } from 'ui-library/src/theme/constants';
```
th0rOdinson, line 19: This function should not ve available like this, we should use it through the theme

th0rOdinson (approved):

apps/root/src/pages/strategy-guardian-detail/vault-data/components/data-guardian/index.tsx
FiboApe: We actually use it quite a bit. Its good when we dont want to introduce a hook to a component since the spacing is going to be constant
Screenshot 2024-08-23 at 2 00 52 PM

PR #956, opened 2024-08-27
fix(*): adjust max-widths

th0rOdinson:
https://github.com/user-attachments/assets/9d59c5ed-e099-40eb-993c-5faab5b00996


FiboApe (approved):
Perfect!

PR #968, opened 2024-08-29
fix(address): set min height of address component so it doesnt flicke…

FiboApe:
…r on hover


apps/root/src/common/components/address/index.tsx
```diff
@@ -9,6 +9,7 @@ import { NETWORKS } from '@constants';
 import { defineMessage, useIntl } from 'react-intl';
 import EditLabelInput from '../edit-label-input';
 import useWallets from '@hooks/useWallets';
+import { SPACING } from 'ui-library/src/theme/constants';
```
th0rOdinson, line 12: We shouldn't import directly from sub-folders

th0rOdinson (approved):

th0rOdinson (approved):

PR #988, opened 2024-09-11
restore avalanche dca banner

th0rOdinson:
Screenshot 2024-09-11 at 10 23 31 AM

### DIFF
--- a/apps/root/src/common/components/news-banner/index.tsx
+++ b/apps/root/src/common/components/news-banner/index.tsx
@@ -1,9 +1,9 @@
-// import { NETWORKS } from '@constants';
-// import { DCA_CREATE_ROUTE } from '@constants/routes';
-// import usePushToHistory from '@hooks/usePushToHistory';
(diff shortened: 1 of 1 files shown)


apps/root/src/common/components/news-banner/index.tsx
```diff
@@ -1,9 +1,9 @@
-// import { NETWORKS } from '@constants';
-// import { DCA_CREATE_ROUTE } from '@constants/routes';
-// import usePushToHistory from '@hooks/usePushToHistory';
+import { NETWORKS } from '@constants';
```
FiboApe, line 1: Should we have different components for each banner and just change which one renders? so we dont have to keep commenting/uncommenting stuff on this file

th0rOdinson: I think we could accumulate a lot of files doing so creating one component for each campaing. But I managed to abstract the parameters and it's more usable now for DX

FiboApe (approved):

PR #994, opened 2024-09-12
fix(strategy-detail): showing day earnings

th0rOdinson:


apps/root/src/pages/strategy-guardian-detail/investment-data/components/financial-data/index.tsx
```diff
@@ -23,7 +22,7 @@ const FinancialData = ({ strategy }: FinancialDataProps) => {
             description="strategy-detail.vault-investment-data.expected-returns"
           />
         </Typography>
-        <ExpectedReturns userPositions={strategy.userPositions} hidePeriods={[StrategyReturnPeriods.DAY]} />
```
FiboApe, line 26: We should have the same for deposit

FiboApe (approved):

PR #995, opened 2024-09-13
fix(timeline): Changing layout structure + other fixes

th0rOdinson:


FiboApe (commented):
Seems like the icons are not the correct one, same for the link icon.

Could we put the time in the time format the ux designer has sent if it is > 1 month?

FiboApe (approved):

PR #998, opened 2024-09-16
feat(design-system): Spacing

th0rOdinson:


packages/ui-library/src/components/dashboard/index.tsx
```diff
@@ -10,8 +10,8 @@ import { formatUsdAmount } from '../../common/utils/currency';
 import { useIntl } from 'react-intl';
 
 const StyledBackgroundPaper = styled(BackgroundPaper)`
-  ${({ theme: { spacing, palette } }) => `
-    padding: ${spacing(5)};
+  ${({ theme: { space, palette } }) => `
+    padding: ${space.s05};
```
FiboApe, line 14: nice! i like it

apps/root/src/pages/transfer/components/transfer-form/index.tsx
```diff
@@ -44,6 +44,10 @@ import FormWalletSelector from '@common/components/form-wallet-selector';
 
 const StyledTransferForm = styled(BackgroundPaper)`
   position: relative;
+  ${({ theme: { space } }) => `
```
FiboApe, line 47: Shouldn't this be integrated to the background paper? Seems like there are already rules into the spacing it should have

FiboApe:
Some of the spacings in the activity are a bit off too and in token profile

FiboApe (approved):

PR #1009, opened 2024-09-19
fix(walletclients): set selected network to actual network of provide…

FiboApe:
…r and set active wallet on link


apps/root/src/pages/transfer/components/transfer-form/index.tsx
```diff
@@ -144,6 +146,12 @@ const TransferForm = () => {
     }
   }, []);
 
+  React.useEffect(() => {
+    if (selectedNetwork.chainId !== actualCurrentNetwork.chainId && !chainIdParam) {
+      dispatch(setChainId(actualCurrentNetwork.chainId));
+    }
+  }, [actualCurrentNetwork]);
+
```
th0rOdinson, line 154: We are having two useEffect in every app, related to setting the network. Would it be possible to merge those? I'm not sure if they will execute in the order we want every time

th0rOdinson (approved):

PR #1011, opened 2024-09-19
fix(earn): fix how we use apy

FiboApe:


th0rOdinson (commented):
Can we include here a check for how we use guardian fees?

apps/root/src/common/utils/earn/parsing.tsx
```diff
@@ -387,3 +388,21 @@ export function calculateEarnFeeAmount({
...
+  feeType: FeeType;
+  assetAmount?: bigint;
+}) {
+  const feePercentage = strategy?.guardian?.fees.find((fee) => fee.type === feeType)?.percentage;
+
+  if (!feePercentage || isUndefined(assetAmount)) return undefined;
+
+  const feePercentageBigInt = BigInt(Math.round(feePercentage * 100));
+
+  return (assetAmount * feePercentageBigInt) / 100000n;
```
th0rOdinson, line 407: A big similar to the function above 😅

th0rOdinson (approved):

PR #1024, opened 2024-09-27
fix(walletconnect): do not ask for wallet client on page load since i…

FiboApe:
…t triggers walletconnect loop


apps/root/src/services/walletClientsService.ts
```diff
@@ -121,7 +104,7 @@ export default class WalletClientsService extends EventsManager<WalletClientsSer
           status: isConnected ? 'connected' : 'disconnected',
           walletClient: client,
```
th0rOdinson, line 105: client would be always be undefined here right?

th0rOdinson (approved):

PR #1026, opened 2024-09-27
fix(sdk): updagrading to 0.0.19

th0rOdinson:


package.json
```diff
@@ -36,7 +36,7 @@
     "turbo": "latest"
   },
   "resolutions": {
-    "@balmy/sdk": "npm:@koaj/earn-sdk@0.0.17"
+    "@balmy/sdk": "npm:@koaj/earn-sdk@0.0.19"
```
FiboApe, line 39: common should be updated too, and check if the root package.json isnt using the koaj one

FiboApe (approved):

PR #1028, opened 2024-09-30
feat(design-system): option picker input

th0rOdinson:


apps/root/src/pages/aggregator/swap-container/components/step1/index.tsx
```diff
@@ -100,7 +100,6 @@ const SwapFirstStep = ({
     : (selectedRoute &&
         selectedRoute?.buyToken.address === to?.address &&
         formatUnits(selectedRoute.buyAmount.amount, selectedRoute.buyToken.decimals || 18)) ||
-      '0.0' ||
```
FiboApe, line 103: we want to show the 0.0 on the inputs right?

packages/ui-library/src/components/select/index.tsx
```diff
@@ -179,11 +176,8 @@ function Select<T extends { key: string | number }, H = object>({
       }}
     >
       {!disabledSearch && (
-        <ListSubheader disableGutters>
+        <ListSubheader disableGutters sx={{ marginBottom: SPACING(2.5) }}>
           <TextField
-            size="small"
-            // Autofocus on textfield
-            autoFocus
```
FiboApe, line 186: we definitely want this

packages/ui-library/src/components/token-picker/index.tsx
```diff
@@ -176,8 +176,8 @@ const Row: ItemContent<TokenWithBalance, RowData> = (
     <StyledForegroundPaper key={`${token.chainId}-${token.address}`} onClick={() => onClick(tokenWithBalance)}>
       <ContainerBox flex="1" alignItems="center" gap={3}>
         {token.icon}
-        <ContainerBox flexDirection="column" flex="1" alignItems="flex-start">
-          <Typography variant="bodyBold" color={colors[themeMode].typography.typo2}>
+        <ContainerBox flexDirection="column" flex="1" alignItems="flex-start" gap={0.5}>
+          <Typography variant="bodyBold" color={colors[themeMode].typography.typo2} lineHeight={1}>
```
FiboApe, line 180: this has a specific different line height?

packages/ui-library/src/components/virtualized-list/index.tsx
```diff
@@ -1,37 +1,37 @@
 import React, { forwardRef } from 'react';
-import { Components, ItemContent, Virtuoso } from 'react-virtuoso';
+import { ItemContent, Virtuoso } from 'react-virtuoso';
 import styled from 'styled-components';
 
-const StyledList = styled.div`
+const StyledList = styled.div<{ $gap?: number }>`
```
FiboApe, line 5: This should just be a containerBox

packages/ui-library/src/components/virtualized-list/index.tsx
```diff
@@ -1,37 +1,37 @@
...
+const VirtualizedList = <D, C>({ data, itemContent, fetchMore, context, gap }: VirtualizedListProps<D, C>) => (
   <Virtuoso
     style={{ height: '100%' }}
     data={data}
     itemContent={itemContent}
     endReached={fetchMore}
     context={context}
-    components={VirtuosoComponents}
+    components={{
+      List: forwardRef<HTMLDivElement>(function VirtuosoList(props, ref) {
```
FiboApe, line 31: If we create it outside it doesnt get created on every re-render and saves up on memory

FiboApe:
Screenshot 2024-09-30 at 1 40 52 pm
Paddings of the contacts modal are wrong

Also the contacts cards are missing their hover styles, same for thje options menu options

FiboApe:
Screenshot 2024-09-30 at 1 43 40 pm
Also some logos are having some weird box-shadow, maybe we should roujnd them

packages/ui-library/src/components/select/index.tsx
th0rOdinson: This was suggested by the designer in the notion, I'll undo it anyway!

packages/ui-library/src/components/token-picker/index.tsx
th0rOdinson: By default it's line height is 2, so in order to make the line thiner I made that change. With default line height:
Screenshot 2024-09-30 at 1 54 42 PM

FiboApe (approved):

PR #1035, opened 2024-10-02
feat(design-system): review buttons definitions

FiboApe:


th0rOdinson (commented):
Great job! Some differences I've found with the design:
- On home/portfolio with no wallet connected:
  - Main connect wallet button is much wider, should not be fullwidth
  - Connect wallet buttons from both wallet selectors (main and nav) are in color primary, that looks odd. Also, it has too much height
- In history table, the 'View More' receipt button should all be in color primary, and a little bolder too. This button also adds too much height to each row
- Earn Strategies Table Toolbar's buttons have too much height, breaking the design for that container
Screenshot 2024-10-03 at 11 40 34 AM

 - Minimal Buttons doesnt match the design in Historical APY for strategies detail. Also, when hovering them, their size changes
Screenshot 2024-10-03 at 11 41 47 AM

- This also happens in token profile chart
- About history timeline, tabs now have too much height, and maybe some horizontal padding. Also, they should be more rounded
- Hover usually changes the size of the outline button, this is visible too in dca position card button
- Withdraw button in Dca position details is also too tall:
Screenshot 2024-10-03 at 11 55 05 AM

- Transaction confirmation now is a little thiner, it should use button max width
Screenshot 2024-10-03 at 12 00 09 PM

th0rOdinson (commented):
- Home with no wallet connected still looks odd. I think it should not be color primary. Plus, connect your wallet button should not be fullwidth inside portfolio paper as mentioned before
Screenshot 2024-10-04 at 12 12 11 PM

- Earn table filters still doesnt match search input height:
Screenshot 2024-10-04 at 12 15 11 PM

- Period buttons in graph container look much better, but now they have an unnecessary shadow
Screenshot 2024-10-04 at 12 17 05 PM
- I think navigations pills should use quarteryNoAlpha as background color

th0rOdinson (approved):

PR #1037, opened 2024-10-03
feat(design-system): tables

th0rOdinson:


packages/ui-library/src/components/sort-icon/index.tsx
```diff
@@ -0,0 +1,26 @@
...
+
+const SortIcon = ({ direction }: SortIconProps) => {
+  const {
+    palette: { mode },
+  } = useTheme();
+
+  const arrowUpColor = direction === 'asc' ? colors[mode].typography.typo1 : colors[mode].typography.typo5;
+  const arrowDownColor = direction === 'desc' ? colors[mode].typography.typo1 : colors[mode].typography.typo5;
+  return (
+    <ContainerBox flexDirection="column" alignItems="center">
```
FiboApe, line 19: Shouldn't this actually be just 1 svg? Or is it made this way in figma?

FiboApe: Ah gotcha need different colors depending on direction

FiboApe:
- Earn Strategy details, the breakdown shows the background is different:
Screenshot 2024-10-04 at 2 38 51 pm
- Also the spacing is off (see that it is applying 8px two times after the divider
Screenshot 2024-10-04 at 2 41 05 pm
- Pagination buttons do not match the figma
Screenshot 2024-10-04 at 2 42 48 pm
- Portfolio table should not be paginated
Screenshot 2024-10-04 at 2 43 49 pm
- Alignments on the porfolio table do not match those in figma (also on the token profile page)
Screenshot 2024-10-04 at 2 45 19 pm

packages/ui-library/src/theme/typography.ts
```diff
@@ -197,4 +197,12 @@ export const buildTypographyVariant = (mode: PaletteMode): TypographyOptions =>
     lineHeight: 1.33, // 16/12
     color: colors[mode].typography.typo3,
   },
+  tablePaginationLabel: {
+    fontFamily: 'Archivo',
```
FiboApe, line 201: we should not be using other fontFamilies, check with designers if they have missed this, but it should all be Inter always

FiboApe (approved):

PR #1038, opened 2024-10-04
feat(earn): handle delay withdraw status

th0rOdinson:


apps/root/src/common/utils/earn/parsing.tsx
```diff
@@ -28,6 +30,16 @@ import { TableStrategy } from '@pages/earn/components/strategies-table';
 import { ColumnOrder, StrategiesTableVariants } from '@state/strategies-filters/reducer';
 import { Address, formatUnits, parseUnits } from 'viem';
 
+export const sdkStrategyTokenToTokenWithWitdrawTypes = (
+  sdkToken: SdkStrategyTokenWithWithdrawTypes,
+  tokenKey: TokenListId,
+  tokenList: TokenList,
+  chainId?: number
+): TokenWithWitdrawTypes => {
+  const token = tokenList[tokenKey] || toToken({ ...sdkToken, chainId });
```
FiboApe, line 39: couldn't this just call sdkStrategyTokenToToken and add the withdraw type?

apps/root/src/frame/polling-handlers/earn-positions-updater/index.tsx
```diff
@@ -0,0 +1,19 @@
...
+import useInterval from '@hooks/useInterval';
+import useEarnService from '@hooks/earn/useEarnService';
+import { IntervalSetActions } from '@constants/timing';
+
+const EarnPositionsUpdater = () => {
+  const earnService = useEarnService();
+  const updateInterval = IntervalSetActions.earnPositionsUpdate;
+
+  const updateBalancesAndPrices = React.useCallback(async () => {
+    await earnService.fetchUserStrategies();
```
FiboApe, line 11: Be carefull since this should extend the positions that we have right now, so for example it would remove the history from the detailed userStrategies

apps/root/src/services/earnService.ts
```diff
@@ -341,11 +365,13 @@ export class EarnService extends EventsManager<EarnServiceData> {
       strategiesArray.map((userStrategy) => userStrategy.strategy)
     );
 
-    this.userStrategies = savedUserStrategies;
+    if (updateStore) {
```
FiboApe, line 368: can't we update them here, so we only have 1 fetchUserStrategies method?, if there are positions, we update them, if there are not, we add them

apps/root/src/services/earnService.ts
```diff
@@ -301,6 +301,15 @@ export class EarnService extends EventsManager<EarnServiceData> {
     this.allStrategies = allStrategies;
   }
 
+  batchUpdateUserStrategies(userStrategies: SdkEarnPosition[]) {
+    let storedUserStrategies = [...this.userStrategies];
+
+    userStrategies.forEach((strategy) => {
+      storedUserStrategies = this.updateUserStrategy(strategy, false);
```
FiboApe, line 308: if its not updating the store, this is only going to apply for the last strategy passed here, since updateUserStrategy is going to extend from this.userStrategies

FiboApe (approved):
Now it looks good

PR #1044, opened 2024-10-10
feat(*): inputs revisions

FiboApe:


apps/root/src/pages/earn/components/strategies-table/components/filters/index.tsx
```diff
@@ -390,12 +389,7 @@ const TableFilters = ({ isLoading, variant }: TableFiltersProps) => {
 
   return (
     <ContainerBox alignItems="center" justifyContent="flex-end" gap={3}>
-      <StyledControlButton
-        onClick={handleOpen}
-        disabled={isLoading}
-        variant="outlined"
-        endIcon={<KeyboardArrowDownIcon />}
-      >
+      <StyledControlButton onClick={handleOpen} disabled={isLoading} variant="outlined">
```
th0rOdinson, line 392: why remove it?

packages/ui-library/src/components/token-amount-usd-input/index.tsx
```diff
@@ -177,6 +202,30 @@ const UsdInput = ({ onChange, value, token, tokenPrice, onBlur, onFocus, disable
...
 };
 
+const StyledHeader = styled(ContainerBox).attrs(() => ({ gap: 3, justifyContent: 'center', alignItems: 'center' }))``;
+
+const StyledFooter = styled(ContainerBox).attrs(() => ({ gap: 3, justifyContent: 'center', alignItems: 'center' }))``;
+
+const StyledEndContent = styled(ContainerBox).attrs(() => ({
+  gap: 3,
+  justifyContent: 'space-between',
+  alignItems: 'center',
```
th0rOdinson, line 212: maybe we should align `end`? This doesn't match the design (DCA Create):
Screenshot 2024-10-16 at 3 24 21 PM

th0rOdinson (commented):
Some differences found by comparing with the design:

- Earn wizard
Screenshot 2024-10-16 at 3 27 31 PM

- Search input in Strategies Table Toolbar match siblings height
Screenshot 2024-10-16 at 3 28 23 PM

- Earn deposit input let's me initiate steps when amount is "0"

Screenshot 2024-10-16 at 3 33 58 PM

- In DCA Investment management, duration input doesn't match height
- Form wallet selector dropdown icon is not well positioned
Screenshot 2024-10-16 at 3 35 36 PM

- I think it would look much better if token amounts doesn't push token pickers:
Screenshot 2024-10-16 at 3 36 15 PM

- As on manage investment, investment duration in DCA create also doesnt match height

- This is a known issue; but in DCA Create, when depositing Protocol token, max button uses total balance, without discounting gas estimations

- In agg settings, we could make this a single row, and also would not match height in that case
Screenshot 2024-10-16 at 3 40 42 PM

packages/ui-library/src/theme/typography.ts
```diff
@@ -205,4 +205,12 @@ export const buildTypographyVariant = (mode: PaletteMode): TypographyOptions =>
     lineHeight: 1.33, // 16/12
     color: colors[mode].typography.typo3,
   },
+  labelSemiBold: {
```
th0rOdinson, line 208: Also this should be label Large if fontSize needs to be 14px

th0rOdinson: Still!

th0rOdinson (approved):

PR #1045, opened 2024-10-10
fix(token-distribution): do not show balances for tokens of earn that…

FiboApe:
… do not apply

### DIFF
--- a/apps/root/src/hooks/useNetWorth.ts
+++ b/apps/root/src/hooks/useNetWorth.ts
@@ -60,85 +60,109 @@ const useNetWorth = ({ walletSelector, chainId, tokens }: NetWorthProps) => {
           });
         return acc;
       }, {}),
-    [allBalances, chainId]
+    [allBalances, chainId, tokens]
   );
-
-  let walletAssetsTotalValue = 0;
 
   const walletAddressToEvaluate =
     walletSelector !== ALL_WALLETS && (walletSelector || activeWallet?.address || authWallet);
 
-  if (walletSelector === ALL_WALLETS) {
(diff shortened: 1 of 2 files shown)


apps/root/src/constants/addresses.ts
```diff
@@ -593,7 +593,7 @@ export const EARN_COMPANION_ADDRESS: Record<number, Address> = {
 };
 
 export const EARN_VAULT_ADDRESS: Record<number, Address> = {
-  [NETWORKS.polygon.chainId]: '0x0aA2c8C6d099bEC1Bef626F45cEF7506b42B4Ad0',
+  [NETWORKS.polygon.chainId]: '0x63a8aE714568EC8f8Ec14472674c68582b0B0458',
```
th0rOdinson, line 596: This change should be there already 🤔

yarn.lock
```diff
@@ -3042,7 +3043,7 @@
   resolved "https://registry.yarnpkg.com/@noble/hashes/-/hashes-1.4.0.tgz#45814aa329f30e4fe0ba49426f49dfccdd066426"
   integrity sha512-V1JJ1WTRUqHHrOSh597hURcMqVKVGL/ea3kv0gSnEdsEZ0/+VyPghM1lMNGc00z7CIQorSvbKpuJkxvuHbvdbg==
 
-"@noble/hashes@1.5.0", "@noble/hashes@^1.4.0":
+"@noble/hashes@1.5.0", "@noble/hashes@^1.4.0", "@noble/hashes@~1.5.0":
```
th0rOdinson, line 3046: This happened to me too when doing yarn install. Not sure if still necessary after merging the sdk update branch

th0rOdinson (approved):

PR #1048, opened 2024-10-11
feat(*): add earn coming soon

FiboApe:


apps/root/src/frame/components/navigation/index.tsx
```diff
@@ -254,6 +282,13 @@ const Navigation = ({ children }: React.PropsWithChildren) => {
           label: intl.formatMessage(DASHBOARD_ROUTE.label),
           type: SectionType.link,
         },
+        {
+          label: intl.formatMessage(defineMessage({ description: 'earn', defaultMessage: 'Earn' })),
+          key: 'earn-group',
```
th0rOdinson, line 287: We should not allow any redirects from clicking this item, it's sending me to /earn-group, which renders /create from DCA

th0rOdinson (approved):

PR #1053, opened 2024-10-14
Bly 3125

FiboApe:


apps/root/src/hooks/earn/useSuggestedStrategies.ts
```diff
@@ -21,7 +21,9 @@ export default function useSuggestedStrategies(
       strategiesWithReward: Strategy[];
     }>(
       (acc, strategy) => {
-        const isSameAsset = getIsSameOrTokenEquivalent(selectedAsset.token, strategy.asset);
+        const isSameAsset =
+          getIsSameOrTokenEquivalent(selectedAsset.token, strategy.asset) ||
+          getIsSameOrTokenEquivalent(selectedReward.token, strategy.asset);
```
th0rOdinson, line 26: This is not happening for now, but there could be strategies that don't generate yield in their asset token. We can check that by seeing the APY

th0rOdinson (approved):

PR #1057, opened 2024-10-15
fix(earn): integrate with new earn api event types

th0rOdinson:


apps/root/src/common/utils/transaction-history/index.ts
```diff
@@ -188,20 +188,21 @@ export const getTransactionValue = (tx: TransactionEvent, intl: ReturnType<typeo
...
     case TransactionEventTypes.EARN_CREATED:
     case TransactionEventTypes.EARN_INCREASE:
-      return `${formatCurrencyAmount({ amount: tx.data.assetAmount.amount, token: tx.data.asset, intl })} ${
-        tx.data.asset.symbol
+      return `${formatCurrencyAmount({ amount: tx.data.depositAmount.amount, token: tx.data.depositToken, intl })} ${
+        tx.data.depositToken.symbol
       }`;
     case TransactionEventTypes.EARN_WITHDRAW:
-      const assetAmount = tx.data.withdrawn.find((withdrawn) => withdrawn.token.address === tx.data.assetAddress);
+      const withdrawnAsset = tx.data.withdrawn[0];
```
FiboApe, line 195: why only the first one?

packages/common-types/src/accountHistory.ts
```diff
@@ -163,42 +162,37 @@ export interface SwapApiEvent {
...
 
 export interface EarnDepositApiDataEvent {
-  strategyId: StrategyId;
-  positionId: SdkEarnPositionId;
-  asset: {
-    amount: string;
-    address: Address;
-    price?: number;
-  };
+  owner: Address;
```
FiboApe, line 165: can't we also have the strategyId so we can link to it in the receipt?

FiboApe: Same comment about all other events

packages/common-types/src/accountHistory.ts
```diff
@@ -163,42 +162,37 @@ export interface SwapApiEvent {
...
 export interface EarnIncreaseApiDataEvent {
-  strategyId: StrategyId;
-  positionId: SdkEarnPositionId;
-  asset: {
-    amount: string;
-    address: Address;
-    price?: number;
-  };
+  assetPrice?: number;
+  assetsDeposited: string;
```
FiboApe, line 179: can it be several or is it only one?

apps/root/src/pages/history/components/historyTable/index.tsx
```diff
@@ -209,29 +209,35 @@ const formatTokenElement = (txEvent: TransactionEvent): React.ReactElement => {
...
+              {txEvent.data.depositToken.name || '-'}
             </StyledBodySmallLabelTypography>
           </StyledCellContainer>
         </>
       );
     case TransactionEventTypes.EARN_WITHDRAW:
-      const tokens = txEvent.data.withdrawn.map((withdrawn) => withdrawn.token);
-      const assetToken = tokens.find((token) => token.address === txEvent.data.assetAddress);
-      const isWithdrawingRewards = tokens.some((token) => token.address !== txEvent.data.assetAddress);
+      const withdrawnAsset = txEvent.data.withdrawn[0];
```
FiboApe, line 224: this actually seems a bit random that it will always be the first, what happens if they later change that?

apps/root/src/common/utils/transaction-history/index.ts
th0rOdinson: It's a convention from the BE. They will always return all tokens, the asset first, and the amount = 0 will tell us which one's are withdrawn

packages/common-types/src/accountHistory.ts
th0rOdinson: I have asked for this data, do you think it's a priority right now? I think we might need it for the release, but not as an urgent

apps/root/src/pages/history/components/historyTable/index.tsx
th0rOdinson: I'm asking for the assetAddress anyway so we can get the transactionValue in usd. So I think this will change in the next commit. Anyway it's a convention as mentioned before, but I agree it's a little sensitive for future changes

packages/common-types/src/accountHistory.ts
th0rOdinson: Only one. It's the amount expressed in asset token

packages/common-types/src/accountHistory.ts
FiboApe: Yes for the release, not for now

packages/common-types/src/accountHistory.ts
FiboApe: then it should be assetDepositedAmount :)

th0rOdinson: They use that naming for their contracts. I will change it for our parsed events, but the api would not change from their side

FiboApe: then don't change it on our side

FiboApe (approved):

PR #1062, opened 2024-10-17
feat(delay-withdraw): Add delay info to strategy management

th0rOdinson:


apps/root/src/services/earnService.ts
```diff
@@ -1019,6 +1024,52 @@ export class EarnService extends EventsManager<EarnServiceData> {
           timestamp: nowInSeconds(),
         });
 
+        // Update Delayed Withdraw Data
+        const assetWithdrew = withdrawn.find((w) => w.token.address === strategy.farm.asset.address);
```
FiboApe, line 1028: we should be using isSameToken right?

apps/root/src/services/earnService.ts
```diff
@@ -1037,11 +1088,7 @@ export class EarnService extends EventsManager<EarnServiceData> {
 
         modifiedStrategy.pendingTransaction = '';
 
-        const userHasRemainingFunds = modifiedStrategy.balances.some((balance) => balance.amount.amount > 0n);
-
-        if (userHasRemainingFunds) {
-          userStrategies.push(modifiedStrategy);
-        }
+        userStrategies.push(modifiedStrategy);
```
FiboApe, line 1091: good, we need to handle then to not show positions that dont have funds, but this is for another ticket

apps/root/src/services/earnService.spec.ts
```diff
@@ -714,13 +716,118 @@ describe('Earn Service', () => {
...
+                  withdrawType: WithdrawType.IMMEDIATE,
                 },
               ],
               positionId: '10-0xvault-30',
               strategyId: '0xvault' as `${number}-${Lowercase<string>}-${number}`,
             } satisfies EarnWithdrawTypeData['typeData'],
           },
         },
+        {
+          expectedPositionChanges: {
```
FiboApe, line 728: can we also add a case for one that did not have delayed before?

apps/root/src/pages/strategy-guardian-detail/strategy-management/withdraw/tx-manager/index.tsx
```diff
@@ -37,8 +42,41 @@ const EarnWithdrawTransactionManager = ({ strategy, setHeight }: EarnWithdrawTra
 
   return (
     <>
+      <Modal
```
FiboApe, line 45: This should be on a different component

apps/root/src/pages/strategy-guardian-detail/strategy-management/withdraw/tx-manager/index.tsx
```diff
@@ -37,8 +42,41 @@ const EarnWithdrawTransactionManager = ({ strategy, setHeight }: EarnWithdrawTra
...
+          <FormattedMessage
+            defaultMessage="You're about to make an instant withdrawal from your vault. Fees will be applied to your total withdrawal, allowing you to skip the standard waiting period."
+            description="earn.strategy-management.market-withdrawal-modal.subtitle"
+          />
+        }
+        maxWidth="sm"
+      >
+        <ContainerBox gap={4}>
+          <Button variant="outlined" size="large" onClick={onMarketWithdrawCancel} fullWidth>
+            <FormattedMessage defaultMessage="Cancel" description="cancel" />
```
FiboApe, line 66: this should go as actions on the modal

apps/root/src/pages/strategy-guardian-detail/strategy-management/withdraw/tx-manager/index.tsx
FiboApe: Also isnt this modal missing the expected changes?

apps/root/src/common/components/transaction-steps/index.tsx
```diff
@@ -824,7 +820,8 @@ const buildEarnWithdrawItem = ({
     >
       {isCurrentStep && (
         <StyledTransactionStepButtonContainer>
-          <Button variant="contained" fullWidth size="large" onClick={() => onAction(transactions)}>
+          {/* // TODO: Check if market withdraw would go by txSteps if companion has no withdraw permission */}
+          <Button variant="contained" fullWidth size="large" onClick={() => onAction(WithdrawType.IMMEDIATE)}>
```
FiboApe, line 824: should'nt it be market if its delayed?

apps/root/src/common/utils/currency.ts
```diff
@@ -6,7 +6,7 @@ import _Decimal from 'decimal.js-light';
 import JSBI from 'jsbi';
 import toFormat from 'toformat';
 import { NetworkStruct, Token, TokenType } from '@types';
-import { DCAPositionToken, TokenVariant } from '@balmy/sdk';
+import { DCAPositionToken, isSameAddress, TokenVariant } from '@balmy/sdk';
```
FiboApe, line 9: this import is needed?

apps/root/src/pages/strategy-guardian-detail/strategy-management/components/delay-withdraw-warning/index.tsx
```diff
@@ -0,0 +1,50 @@
...
+    border-radius: 50%;
+    background-color: ${colors[palette.mode].background.tertiary};
+  `}
+`;
+
+const DelayWithdrawWarning = () => (
+  <StyledDelayWithdrawWarningContainer gap={1} alignItems="center">
+    <ButtonIconContainer>
+      <TimerIcon
+        sx={({ spacing }) => ({
```
FiboApe, line 24: we could just send fontSize as the prop, since it already handles that the svg icon

apps/root/src/pages/strategy-guardian-detail/strategy-management/components/delay-withdraw-warning/index.tsx
```diff
@@ -0,0 +1,50 @@
...
+  ${({ theme: { palette, spacing } }) => `
+    padding: ${spacing(1.125)};
+    border-radius: 50%;
+    background-color: ${colors[palette.mode].background.tertiary};
+  `}
+`;
+
+const DelayWithdrawWarning = () => (
+  <StyledDelayWithdrawWarningContainer gap={1} alignItems="center">
+    <ButtonIconContainer>
```
FiboApe, line 22: Why is this a buttonIconContainer btw?

apps/root/src/pages/strategy-guardian-detail/strategy-management/index.tsx
```diff
@@ -35,25 +37,30 @@ const StrategyManagement = ({ chainId, strategyGuardianId }: StrategyManagementP
...
-        />
-        <Tab
-          label={
-            <Typography variant="bodyRegular" color="inherit">
-              <FormattedMessage description="earn.strategy-management.tabs.withdraw" defaultMessage="Withdraw" />
-            </Typography>
-          }
-        />
-      </UnderlinedTabs>
+      <ContainerBox>
```
FiboApe, line 45: what about gaps and alignment?

apps/root/src/pages/strategy-guardian-detail/strategy-management/withdraw/cta-button/index.tsx
```diff
@@ -153,12 +164,49 @@ const EarnWithdrawCTAButton = ({ strategy, onHandleWithdraw, onHandleProceed }:
...
+          defaultMessage="Initiate Withdraw"
+        />
+      </Button>
+      {shouldShowMarketWithdraw && (
+        <Button
+          size="large"
+          variant="outlined"
+          disabled={!!shouldDisableProceedButton}
+          fullWidth
+          onClick={companionHasPermission ? () => onWithdraw(WithdrawType.MARKET) : onShowMarketWithdrawModal}
```
FiboApe, line 199: we should always show the modal since it explains what is going to happen

apps/root/src/services/earnService.ts
th0rOdinson: Agree! I have created it here:
https://linear.app/balmy/issue/BLY-3234/filter-out-strategies-with-no-balance-from-portoflio-table

apps/root/src/services/earnService.spec.ts
th0rOdinson: I think this was the case, but I have added one that had some delayed before, so checks the sum :)

apps/root/src/pages/strategy-guardian-detail/strategy-management/withdraw/tx-manager/index.tsx
th0rOdinson: Yes, the modal should do an sdk call to calculate the market swap. For this task i did a quick modal to make the flow work, and I have a task specifically for tackle this. Sorry didn't mentioned it!

apps/root/src/pages/strategy-guardian-detail/strategy-management/withdraw/tx-manager/index.tsx
th0rOdinson: Agree, will tackle in Modal Task

apps/root/src/pages/strategy-guardian-detail/strategy-management/components/delay-withdraw-warning/index.tsx
th0rOdinson: fontSize is theme typed, so custom values (like 15px in this case) would not be accepted

apps/root/src/pages/strategy-guardian-detail/strategy-management/components/delay-withdraw-warning/index.tsx
th0rOdinson: This gives the gray background circle!

apps/root/src/pages/strategy-guardian-detail/strategy-management/index.tsx
th0rOdinson: I think no gap or alignment is needed here. Gap = 0 ensures the underlined from tabs continue with the borderBottom of this warning, and alignItems = stretch ensures that the height of both siblings is the same (and gets this line continuity)

apps/root/src/services/earnService.ts
```diff
@@ -1019,6 +1024,54 @@ export class EarnService extends EventsManager<EarnServiceData> {
...
+        const assetWithdrew = withdrawn.find((w) =>
+          isSameToken(w.token, toToken({ ...strategy.farm.asset, chainId: strategy.farm.chainId }))
+        );
+        const hasInitiatedDelayedWithdraw =
+          assetWithdrew && BigInt(assetWithdrew.amount) > 0n && assetWithdrew.withdrawType === WithdrawType.DELAYED;
+
+        if (hasInitiatedDelayedWithdraw) {
+          modifiedStrategy.delayed = modifiedStrategy.delayed || [];
+          const prevAssetDelayedIndex = modifiedStrategy.delayed.findIndex(
+            (d) => d.token.address === strategy.farm.asset.address
```
FiboApe, line 1037: isSameToken here too :)

FiboApe (approved):

PR #1066, opened 2024-10-18
feat(earn): adjust earn cards ui

th0rOdinson:


apps/root/src/pages/earn/components/strategies-table/components/ready-delayed-withdrawals/index.tsx
```diff
@@ -70,10 +70,10 @@ const ReadyDelayedWithdrawals = () => {
           <>
             {!isActiveWallet ? (
               <Typography variant="bodyExtraSmallBold" color="primary">
-                <Address address={position.owner} />
+                <Address address={position.owner} trimAddress />
```
FiboApe, line 73: trimAddress should actually be default i think, I cant remember a place where we use address and not trim it

FiboApe (approved):

PR #1069, opened 2024-10-20
fix(tx-steps): conform to figma

FiboApe:
Screenshot 2024-10-20 at 1 30 56 am


th0rOdinson (approved):
Looks good!
The paper for Swap Aggregator has too much padding-top when the txSteps are open, we could change that in this task too

PR #1070, opened 2024-10-20
fix(tx-confirmation): do not show dividerBorder1 for last item

FiboApe:


packages/ui-library/src/components/transaction-confirmation/index.tsx
```diff
@@ -238,8 +248,14 @@ const SuccessTransactionConfirmation = ({
...
           <StyledBalanceChangesContainer>
-            {balanceChanges?.map((balanceChange) => (
-              <AmountBalanceChange mode={mode} key={balanceChange.token.address} {...balanceChange} intl={intl} />
+            {balanceChanges?.map((balanceChange, index) => (
+              <AmountBalanceChange
+                mode={mode}
+                key={balanceChange.token.address}
+                {...balanceChange}
+                intl={intl}
+                isLast={index === balanceChanges.length}
```
th0rOdinson, line 257: Wouldn't be index === length - 1?

th0rOdinson (approved):

PR #1081, opened 2024-10-21
fix(earn): dont show empty positions in portfolio

th0rOdinson:


apps/root/src/pages/strategy-guardian-detail/data-frame/index.tsx
```diff
@@ -25,7 +29,7 @@ const VaultDataFrame = ({ chainId, strategyGuardianId }: VaultDataFrameProps) =>
       <Grid item xs={12}>
         <VaultData strategy={strategy} />
```
FiboApe, line 30: vault data has a similar check I think

FiboApe (approved):

PR #1097, opened 2024-10-25
fix(earn): Remove 'all wallets' from wizard selection

th0rOdinson:


packages/ui-library/src/components/select/index.tsx
```diff
@@ -208,9 +208,11 @@ function Select<T extends { key: string | number }, H = object>({
         <ListSubheader
           disableGutters
           sx={{
-            background: colors[mode].background.emphasis,
-            padding: SPACING(3),
-            marginBottom: SPACING(2),
+            border: `1px solid ${colors[mode].accentPrimary}`,
+            background: colors[mode].background.secondary,
+            paddingX: SPACING(3),
+            paddingY: SPACING(3.75),
```
FiboApe, line 214: I know a cursor when I see one 👀 . Lets use padding

FiboApe (approved):

packages/ui-library/src/components/select/index.tsx
th0rOdinson: Haha this was 100% me

PR #1099, opened 2024-10-25
fix(earn): home widgets

th0rOdinson:


apps/root/src/pages/home/components/widget-frame/index.tsx
```diff
@@ -158,9 +165,9 @@ const WidgetFrame = ({
           )}
         </ContainerBox>
         <ContainerBox flex={1}>
-          <Collapse in={shouldShow} sx={{ flex: 1, display: 'flex' }}>
+          <StyledCollapse in={shouldShow} sx={{ flex: 1, display: 'flex' }}>
```
FiboApe, line 168: you dont need this sx now :)

FiboApe (approved):

apps/root/src/pages/home/components/widget-frame/index.tsx
th0rOdinson: True, thanks

PR #1102, opened 2024-10-27
fix(recap-data-deposit): use token with price for recap data

FiboApe:


packages/ui-library/src/theme/variants/svgicon-variants.ts
```diff
@@ -8,7 +8,6 @@ export const buildSvgIconVariant = (mode: 'light' | 'dark'): Components => ({
         color: colors[mode].typography.typo5,
       },
       root: {
-        color: colors[mode].typography.typo3,
```
th0rOdinson, line 11: Most icons are typo3, doesn't this change many?

th0rOdinson (approved):

PR #1103, opened 2024-10-27
feat(activity): change title to revoked if approval is 0

FiboApe:


th0rOdinson (commented):
We should change it in Transaction receipt too

th0rOdinson (approved):

PR #1113, opened 2024-10-30
feat(earn): Allow claiming a delayed withdraw

th0rOdinson:


packages/ui-library/src/components/transaction-receipt/index.tsx
```diff
@@ -619,26 +620,78 @@ const EarnWithdrawTransactionReceipt = ({ transaction }: { transaction: EarnWith
   const { spacing } = useTheme();
   const intl = useIntl();
 
+  return (
+    <>
+      <StyledSectionContent>
+        <Typography variant="labelRegular">
+          <FormattedMessage description="TransactionReceipt-transactionEarnWithdraw" defaultMessage="Withdrew" />
```
FiboApe, line 627: wait are we parsing all sending of delayed withdraw and normal withdraws with the same receipt? Or does this only apply if you also withdrew some token through market?

apps/root/src/common/utils/transaction-history/parsing.tsx
```diff
@@ -91,15 +93,14 @@ const parseDcaCreatedApiEvent: ParseFunction<DCACreatedApiEvent, DCACreatedEvent
...
-        amountInUSD:
-          event.data.fromToken.price === null || isUndefined(event.data.fromToken.price)
-            ? undefined
-            : parseFloat(
-                formatUnits(
-                  BigInt(event.data.rate) * parseNumberUsdPriceToBigInt(event.data.fromToken.price),
-                  dcaBaseEventData.fromToken.decimals + 18
-                )
-              ).toFixed(2),
+        amountInUSD: isNil(event.data.fromToken.price)
```
FiboApe, line 96: why was this change needed? isNil is going to return false if it is undefined as well

apps/root/src/pages/strategy-guardian-detail/strategy-management/withdraw/delayed-withdraw-container/index.tsx
```diff
@@ -124,14 +130,31 @@ const DelayedWithdrawItem = ({ delayed, type, position }: DelayedWithdrawItemPro
...
-      {type === DelayedWithdrawalStatus.READY && (
-        <Button variant="outlined">
-          <FormattedMessage
-            description="earn.strategy-management.withdraw.delayed-withdraw.item.withdraw-now"
-            defaultMessage="Withdraw now"
-          />
-        </Button>
-      )}
+      {type === DelayedWithdrawalStatus.READY &&
+        (!!position.pendingTransaction ? (
```
FiboApe, line 134: No need for this since they can submit several delayed withdraws at the same time

apps/root/src/hooks/earn/useEarnClaimDelayedWithdrawAction.tsx
```diff
@@ -0,0 +1,105 @@
...
+      try {
+        if (!claimToken) throw new Error('No claimable token found');
+
+        setModalLoading({
+          content: (
+            <Typography variant="bodyRegular">
+              <FormattedMessage
+                description="earn.strategy-management.claim-delayed-withdraw.modal.loading"
+                defaultMessage="You are now claiming delayed withdraw from {farm}. Time to enjoy the rewards you've cultivated 🎉"
+                values={{ farm: strategy.farm.name }}
```
FiboApe, line 35: lets also add the token symbol here and the amount in case the copy team want to use it

apps/root/src/services/earnService.ts
```diff
@@ -681,6 +681,56 @@ export class EarnService extends EventsManager<EarnServiceData> {
...
+      throw new Error('Could not find userStrategy');
+    }
+
+    const strategy = this.allStrategies.find((s) => s.id === userStrategy.strategy);
+
+    if (!strategy) {
+      throw new Error('Could not find strategy');
+    }
+
+    // NOTE: For the moment we should never expect the permissionSignature.
```
FiboApe, line 705: Like it that we already support it tho :D

packages/ui-library/src/components/transaction-receipt/index.tsx
th0rOdinson: Yes, they are all the same event, with a different withdrawType prop. I've only added the withdraw Type in the delayed receipt, becuase it's the one that doesn't put money immediately.
Should we have three different receipts?

apps/root/src/pages/strategy-guardian-detail/strategy-management/withdraw/delayed-withdraw-container/index.tsx
th0rOdinson: This would only disable that particular button, so they don't submit the same transaction, and they can check status

FiboApe: Ah right since we have only 1 pending and 1 ready by position

packages/ui-library/src/components/transaction-receipt/index.tsx
```diff
@@ -620,47 +620,40 @@ const EarnWithdrawTransactionReceipt = ({ transaction }: { transaction: EarnWith
   const { spacing } = useTheme();
   const intl = useIntl();
 
+  const isDelayed = transaction.data.withdrawn.some(
```
FiboApe, line 623: can we move this to a util function? already seen it used 3 times

FiboApe (approved):

PR #1115, opened 2024-10-31
fix(*): handle usd amounts differenciated for wallet and network sele…

th0rOdinson:
…ctors

### DIFF
--- a/apps/root/src/state/balances/hooks.ts
+++ b/apps/root/src/state/balances/hooks.ts
@@ -47,7 +47,7 @@ export function useWalletBalances(
   return { balances: tokenBalances, isLoadingBalances: isLoadingAllBalances, isLoadingPrices: isLoadingChainPrices };
 }
 
-export function useUsdBalances(filter?: { chainId: number; tokens: Token[] }): {
+export function useUsdBalances(filter?: { chainId?: number; tokens?: Token[] }): {
   isLoading: boolean;
   usdBalances: AllWalletsBalances;
 } {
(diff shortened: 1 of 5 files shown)


apps/root/src/state/balances/hooks.ts
```diff
@@ -56,37 +56,20 @@ export function useUsdBalances(filter?: { chainId: number; tokens: Token[] }): {
...
-    // No filter, sum up balances across all chains
-    if (!filter) {
-      Object.values(allBalances).forEach((chainBalances) => {
-        Object.values(chainBalances.balancesAndPrices || {}).forEach(({ token, balances, price }) => {
-          Object.entries(balances).forEach(([address, balance]: [Address, bigint]) => {
-            const usdBalance = parseUsdPrice(token, balance, parseNumberUsdPriceToBigInt(price));
-            usdBalances[address] = (usdBalances[address] || 0) + usdBalance;
-          });
+    Object.entries(allBalances).forEach(([chainId, chainBalances]) => {
+      if (filter && Number(chainId) !== filter.chainId) return;
```
FiboApe, line 60: this should also check that filter.chainId is not undefined or null

apps/root/src/state/balances/hooks.ts
```diff
@@ -56,37 +56,20 @@ export function useUsdBalances(filter?: { chainId: number; tokens: Token[] }): {
...
+      Object.values(chainBalances.balancesAndPrices || {}).forEach(({ token, balances, price }) => {
+        if (filter && !filter.tokens.some((t) => isSameToken(t, token))) return;
```
FiboApe, line 62: same comment here about tokens

FiboApe (approved):

PR #1122, opened 2024-11-04
fix(swap): parse wrap and unwrap events

th0rOdinson:


packages/common-types/src/transactions.ts
```diff
@@ -165,7 +149,7 @@ export interface SwapTypeData {
     balanceBefore: string | null;
     transferTo?: string | null;
     swapContract: string;
-    type: 'buy' | 'sell';
+    orderType: 'buy' | 'sell';
```
th0rOdinson, line 152: By removing wrap and unwrap, this type was having conflict with the transaction type itself. By changing this prop name, the issue was solved

FiboApe (approved):

PR #1124, opened 2024-11-05
fix(swap): minimum received symbol and tooltip

th0rOdinson:


apps/root/src/pages/aggregator/swap-container/components/quote-data/index.tsx
```diff
@@ -74,19 +74,22 @@ const QuoteData = ({ quote, to, isBuyOrder }: QuoteDataProps) => {
...
               </Typography>
+              <Tooltip
+                title={
+                  <FormattedMessage
+                    description="quoteDataMaximumTooltip"
+                    defaultMessage="This is the maximum you will spend based on your slippage settings"
+                  />
+                }
+              >
+                <ContainerBox>
```
FiboApe, line 86: do we actually need this container box?

FiboApe (approved):

apps/root/src/pages/aggregator/swap-container/components/quote-data/index.tsx
th0rOdinson: Yes, tooltips doesn't work directly over svg icons, I think becuase they are not capable of holding ref's

PR #1130, opened 2024-11-07
feat(*): allow setting min and max approvals

FiboApe:


apps/root/src/frame/components/navigation/index.tsx
```diff
@@ -357,6 +364,16 @@ const Navigation = ({ children }: React.PropsWithChildren) => {
             type: OptionsMenuOptionType.option,
           },
           ...languageOptions,
+          {
+            label: intl.formatMessage(
+              defineMessage({ description: 'useUnlimitedApproval', defaultMessage: 'Unlimited Approval' })
```
th0rOdinson, line 369: Maybe a tooltip with some info would be useful here?

apps/root/src/common/components/transaction-steps/index.tsx
```diff
@@ -485,27 +479,25 @@ const buildApproveTokenItem = ({
...
-                  variant="outlined"
-                  disabled={isPendingTransaction}
-                >
-                  {specificBtnTextAsSecondary}
-                </Button>
+            <>
+              {isCurrentStep && (
+                <ContainerBox gap={3}>
+                  <Button
+                    onClick={isPermit2Enabled && useUnlimitedApproval ? () => onAction() : () => onAction(amount)}
```
th0rOdinson, line 486: isPermit2Enabled i think comes from agg settings, shouldn't we erase that setting?

FiboApe: Screenshot 2024-11-07 at 5 50 29 pm
We actually use it quite a bit

apps/root/src/frame/components/navigation/index.tsx
FiboApe: We dont actually allow tooltips in this items, in any case, its a feature for more expert users that already know what it means, and for more common users I dont want them even touching it :P

FiboApe: Btw tried with a secondary label and it just looks too big

th0rOdinson (approved):

PR #1135, opened 2024-11-11
fix(token-profile): improve wallet not connected page

FiboApe:


apps/root/src/pages/history/components/historyTable/index.tsx
```diff
@@ -460,20 +460,44 @@ const HistoryTableHeader = () => (
...
-        <FormattedMessage
-          description="noActivityParagraph"
-          defaultMessage="Once you start making transactions, you'll see all your activity here"
-        />
-      </Typography>
+    <StyledCellContainer
+      direction="column"
+      align="center"
+      gap={6}
+      style={{ minHeight: '500px', justifyContent: 'center' }}
```
th0rOdinson, line 485: We can use the height in the actual table that is already a param

th0rOdinson (approved):

PR #1147, opened 2024-11-13
fix(earn): get position id from logs

th0rOdinson:


apps/root/src/services/contractService.ts
```diff
@@ -48,22 +49,33 @@ export default class ContractService {
 
   // ADDRESSES
   getHUBAddress(chainId: number, version?: PositionVersions): Address {
-    return (
-      HUB_ADDRESS[version || LATEST_VERSION][chainId] || HUB_ADDRESS[LATEST_VERSION][chainId]
-    ).toLowerCase() as Address;
+    const versionToUse = version || LATEST_VERSION;
+
+    const hubAddress =
```
FiboApe, line 54: I dont think this should return something if it does not exist on the chain that we asked for, we should instead handle on parseLogs that this can come up as undefined

FiboApe (commented):
This is still going to break parseLog since it will try to do hubCompanionAddress.toLowerCase() and it will be undefined

FiboApe (approved):

PR #1156, opened 2024-11-18
fix(earn-home): logout user correctly from earnService

FiboApe:


apps/root/src/services/earnService.ts
```diff
@@ -170,6 +170,15 @@ export class EarnService extends EventsManager<EarnServiceData> {
...
 
+  logOutUser() {
+    const previousStrategies = this.allStrategies;
+    this.resetData();
+    this.allStrategies = previousStrategies.map((strategy) => ({
+      ...strategy,
+      userPositions: [],
+    }));
+  }
+
```
th0rOdinson, line 181: Ohhh yeah!

th0rOdinson (approved):

PR #1160, opened 2024-11-18
feat(*): add hotjar

FiboApe:


apps/root/src/services/eventService.ts
```diff
@@ -17,6 +18,25 @@ export default class EventService {
...
+        console.error('Error initializing Hotjar', error);
+      }
+    }
+  }
+
+  identifyUser(userId?: string) {
+    if (!userId) {
+      return;
+    }
+    this.mixpanel.identify(userId);
```
th0rOdinson, line 34: We are identifying user twice right? In login method from account service we would trigger it here and in `trackEvent` method

FiboApe: There is no problem with this according to mixpanels docs :)

th0rOdinson (approved):

PR #1163, opened 2024-11-19
fix(earn): fix refresh removing data

FiboApe:


apps/root/src/services/earnService.ts
```diff
@@ -286,6 +286,7 @@ export class EarnService extends EventsManager<EarnServiceData> {
         ...strategy,
         lastUpdatedAt: nowInSeconds(),
         userPositions: includedUserStrategies,
+        hasFetchedHistoricalData: 'hasFetchedHistoricalData' in strategy ? strategy.hasFetchedHistoricalData : false,
```
th0rOdinson, line 289: is this ternary necessary?

apps/root/src/services/earnService.ts
```diff
@@ -374,13 +378,17 @@ export class EarnService extends EventsManager<EarnServiceData> {
 
     return !(
       existingUserStrategy &&
-      'detailed' in existingUserStrategy &&
+      existingUserStrategy.hasFetchedHistory &&
       // eslint-disable-next-line @typescript-eslint/no-unsafe-enum-comparison
-      nowInSeconds() - existingUserStrategy.lastUpdatedAt < IntervalSetActions.strategyUpdate
+      nowInSeconds() - existingUserStrategy.lastUpdatedAt > IntervalSetActions.strategyUpdate
```
th0rOdinson, line 383: oh good catch

th0rOdinson (approved):

apps/root/src/services/earnService.ts
FiboApe: For typescript yes, since this receives either an SdkStrategy or a SavedSdkStrategy

apps/root/src/services/earnService.ts
FiboApe: gotta modify the other one

PR #1171, opened 2024-11-21
fix(earn): DW virtual + claim type error

th0rOdinson:


apps/root/src/pages/strategy-guardian-detail/strategy-management/index.tsx
```diff
@@ -25,6 +25,7 @@ const StyledBackgroundPaper = styled(BackgroundPaper).attrs({ variant: 'outlined
 
 const StyledBadge = styled(Badge)`
   & .MuiBadge-badge {
+    z-index: 0;
```
FiboApe, line 28: should it be 1 for it to always be on top of the button but not the tx confirmation?

FiboApe (approved):

apps/root/src/pages/strategy-guardian-detail/strategy-management/index.tsx
th0rOdinson: default is z-index: 1 actually. The other approach would be to change the z-index for the slide for tx steps and tx confirmation. But for this type of text-button (<Tab> item), I find this simpler as it does the job too

PR #1176, opened 2024-11-24
feat(earn): add env variable for earn enablement

FiboApe:


apps/root/src/frame/polling-handlers/index.tsx
```diff
@@ -10,7 +10,7 @@ const PollingHandlers = () => (
     <TransactionUpdater />
     <BalancesUpdater />
     <NetworkUpdater />
-    <EarnPositionsUpdater />
+    {(process.env.EARN_ENABLED === 'true' && <EarnPositionsUpdater />) ?? null}
```
th0rOdinson, line 13: I would replace `??` with another operator, the first expression will never be null/undefined

apps/root/src/hooks/useNetWorth.ts
```diff
@@ -174,8 +174,9 @@ const useNetWorth = ({ walletSelector, chainId, tokens }: NetWorthProps) => {
     return depositedTokensValue + delayedTokensValue;
   }, [tokens, userStrategies, walletSelector, chainId, walletAddressToEvaluate]);
 
+  const isEarnEnabled = process.env.EARN_ENABLED === 'true';
   const isLoadingSomePrices =
-    (activeWallet && (!hasFetchedCurrentPositions || !hasFetchedUserStrategies)) ||
+    (activeWallet && (!hasFetchedCurrentPositions || (isEarnEnabled && !hasFetchedUserStrategies))) ||
```
th0rOdinson, line 179: How about implementing this condition in the service?

PR #1183, opened 2024-11-26
fix(earn): apply events correctly, by fetching events

th0rOdinson:


apps/root/src/services/sdkService.ts
```diff
@@ -524,6 +524,7 @@ export default class SdkService {
     const positionsByAccount = await this.sdk.earnService.getPositionsByAccount({
       accounts: accounts as ArrayOneOrMore<Address>,
       includeHistoricalBalancesFrom: nowInSeconds() - Number(THREE_MONTHS),
+      includeHistory: true,
```
FiboApe, line 527: we dont need history at this point, why are we adding it?

FiboApe (approved):

PR #1190, opened 2024-11-28
fix(*): Allow changing network when active wallet is not connected

th0rOdinson:


apps/root/src/services/walletService.ts
```diff
@@ -21,7 +21,7 @@ export default class WalletService {
   async changeNetwork(newChainId: number, address?: string, callbackBeforeReload?: () => void): Promise<void> {
     try {
       const currentNetwork = await this.providerService.getNetwork(address);
-      if (currentNetwork.chainId !== newChainId) {
+      if (currentNetwork && currentNetwork.chainId !== newChainId) {
```
FiboApe, line 24:
```suggestion
      if (currentNetwork?.chainId !== newChainId) {
```

Same for the others

FiboApe (approved):

PR #1191, opened 2024-11-29
BLY-3350: Limit height of token profile chart && BLY-3284: Limit width of page titles

th0rOdinson:


apps/root/src/pages/token-profile/market-stats/index.tsx
```diff
@@ -13,12 +13,16 @@ const StyledMarketStatsContainer = styled(BackgroundPaper).attrs({ variant: 'out
...
     padding: ${spacing(6)};
-    gap: ${spacing(10)};
+    gap: ${spacing(8)};
     height: 100%;
     min-height: ${spacing(75)};
     `}
 `;
 
+const StyledBodyBold = styled(Typography).attrs({ variant: 'bodyBold' })`
+  line-height: 1;
```
FiboApe, line 23: why does this need a different lineHeight?

packages/ui-library/src/components/styled/index.tsx
```diff
@@ -27,3 +28,8 @@ export const ColorCircle = styled.div<{ color?: string; size?: number }>`
     display: inline-block
   `}
 `;
+
+export const StyledPageTitleDescription = styled(Typography).attrs({ variant: 'bodyLargeRegular' })`
+  color: ${({ theme: { palette } }) => colors[palette.mode].typography.typo2};
```
FiboApe, line 33: We sure about this one? Lots of those had typo1 defined into it

apps/root/src/pages/token-profile/market-stats/index.tsx
th0rOdinson: default for bodyBold is 2, and it adds too much height to the text and therefore the whole container. Now it can match the design

packages/ui-library/src/components/styled/index.tsx
th0rOdinson: They should all be typo2!

FiboApe (approved):

PR #1196, opened 2024-12-02
fix(earn): update wizard styles

th0rOdinson:


packages/ui-library/src/theme/variants/select-variants.ts
```diff
@@ -30,34 +34,30 @@ export const buildSelectVariant = (mode: 'light' | 'dark'): Components => ({
...
           },
+          '& .Mui-selected': {
+            padding: SPACING(3),
+            backgroundColor: colors[mode].background.emphasis,
+          },
         },
       },
       paper: {
-        backgroundColor: colors[mode].background.secondary,
+        backgroundColor: colors[mode].background.tertiary,
```
th0rOdinson, line 60: In this file we had some duplicated styles, also this bg was not matching design for our selects
This is in prod with bg-color: secondary:
Screenshot 2024-12-02 at 11 02 18 AM

FiboApe (approved):

PR #1197, opened 2024-12-02
feat(permit2addreses): make permit2Addresses more consistent with sdk

FiboApe:


apps/root/src/constants/aggregator.ts
```diff
@@ -120,6 +120,9 @@ export const DEFAULT_AGGREGATOR_SETTINGS: {
   sourceTimeout: TimeoutKey.balance,
 };
 
+const tempSdk = buildSDK();
+const aggSupportedChains = tempSdk.quoteService.supportedChains();
```
th0rOdinson, line 124: If this is aggSupportedChains, can we export this directly? We would erase the manual task of adding chains to the `AGGREGATOR_SUPPORTED_CHAINS: ChainId[]`

th0rOdinson (approved):

0xsambugs (approved):
Great work!

PR #1200, opened 2024-12-03
feat(earn): modal for early access

th0rOdinson:


packages/ui-library/src/components/modal/index.tsx
```diff
@@ -165,7 +169,10 @@ const Modal: React.FC<ModalProps> = ({
       fullWidth
       maxWidth={maxWidth || 'lg'}
       onClose={handleClose}
-      PaperProps={fullHeightProps}
+      PaperProps={{
+        ...fullHeightProps,
+        style: { maxWidth: customMaxWidth },
```
FiboApe, line 174: so which one is going to take precedence? Is there no case of modal that we can use the maxWidth already provided?

apps/root/src/frame/components/earn-gain-access-modal/index.tsx
```diff
@@ -0,0 +1,254 @@
...
+
+  const handleAccessNow = () => {
+    pushToHistory(`/${EARN_ACCESS_NOW_ROUTE.key}`);
+    onClose();
+  };
+
+  const actions = React.useMemo(
+    () =>
+      [
+        ...(!activeWallet?.address && !wallets.length
```
FiboApe, line 142: from what we've seend with the designs, the user would not need to connect here, since we can handle it on the landing page

FiboApe (approved):

PR #1204, opened 2024-12-06
feat(earn): Authorization for early access users + Claim code flow

th0rOdinson:


apps/root/src/frame/index.tsx
```diff
@@ -198,16 +203,7 @@ const AppFrame = ({ config: { wagmiClient } }: AppFrameProps) => {
                               <Route path="/token/:tokenListId" element={<TokenProfile />} />
                               <Route path="/create/:chainId?/:from?/:to?" element={<DCA />} />
                               <Route path="/swap/:chainId?/:from?/:to?" element={<Aggregator />} />
-                              {/* // TODO: Remove this route below it's no longer used (@mixpanel) */}
-                              <Route
-                                path="/:chainId?/:from?/:to?"
-                                element={
```
th0rOdinson, line 204: This route was redirecting users before we were loading account data, so refreshing an /earn route was always getting users outside of earn pages

FiboApe (approved):
Great job! This looks fire !!!

PR #1208, opened 2024-12-11
feat(earn): improve usability for claim code form

th0rOdinson:


apps/root/src/common/components/news-banners/earn-banner/index.tsx
```diff
@@ -44,7 +44,7 @@ const EarnBanner = () => {
 
   return (
     <>
-      {hasEarnAccess ? (
+      {isEarnEnabled ? (
```
th0rOdinson, line 47: Small bug here, when enabling earn env var, the earn-banner would still be showing the suscription banner for `earnEnabled: false` users

FiboApe (approved):

PR #1213, opened 2024-12-18
feat(earn): elegibility criteria functionalities

th0rOdinson:


apps/root/src/pages/earn-access-now/components/elegibility-confirmation/index.tsx
```diff
@@ -74,37 +78,39 @@ const STATUS_MAP: Record<ElegibilityConfirmationStatus, React.ComponentType<Eleg
...
   ElegibilityConfirmationStatus.NEEDS_SIGNATURE,
 ];
 
 const ElegibilityConfirmation = () => {
-  const [status, setStatus] = React.useState(ElegibilityConfirmationStatus.LOADING);
-  const wallets = useWallets();
-  const elegibleWallets: Address[] = wallets.map((wallet) => wallet.address);
+  const { fetchElegibilityAchievements, elegibilityStatus, setElegibilityStatus, elegibleWallets } =
+    useElegibilityCriteria();
+  const wallets = useWallets().map((wallet) => wallet.address);
```
FiboApe, line 94: map is always going to create a new object, it would be better to parse this map on the useEffect directlyt

FiboApe (approved):

PR #1214, opened 2024-12-19
feat(wallet-selector): Wallet verification

th0rOdinson:


apps/root/src/common/components/wallet-selector/useWalletSelectorState.tsx
```diff
@@ -407,18 +455,26 @@ const useWalletSelectorState = ({ options, showWalletCounter }: WalletSelectorBa
...
+        <ContainerBox alignItems="center" gap={1}>
+          <Typography variant="bodySmallBold" color={({ palette: { mode } }) => colors[mode].typography.typo2}>
+            {intl.formatMessage(
+              defineMessage({
+                defaultMessage: 'All',
+                description: 'allWallets',
+              })
+            )}
+          </Typography>
+          {allWalletsVerified ? (
```
FiboApe, line 468: This is not just if all wallets are verified, but if you have a wallet selected and it is verified it should also apply the success icon

FiboApe (approved):

PR #1215, opened 2024-12-19
refactor(tierservice): refactor tier calculation

FiboApe:


apps/root/src/services/tierService.ts
```diff
@@ -15,49 +15,74 @@ export interface TierServiceData {
...
+    requirements: [
+      {
+        type: 'AND',
+        requirements: [
+          { id: AchievementKeys.REFERRALS, value: 3 },
+          {
+            type: 'OR',
+            requirements: [
+              { id: AchievementKeys.SWAP_VOLUME, value: 1000 },
+              { id: AchievementKeys.TWEET, value: 1 },
```
0xsambugs, line 64: I think the API returns true/false here, not a 0/1 value

0xsambugs (approved):
Logic seems ok 👍

th0rOdinson (approved):
Nice testing here!

PR #1227, opened 2025-01-06
feat: implement profile tracking

0xged:


apps/root/src/services/analyticsService.ts
```diff
@@ -0,0 +1,421 @@
...
+      },
+      dca: {
+        total_invested_usd: -Number(usdValueDiff),
+      },
+    });
+  }
+
+  trackSlippageChanged({ slippage }: { slippage: string }) {
+    this.setPeopleProperty({
+      swap: {
```
th0rOdinson, line 176: For this agg config, shoudn't we set also the last_product_used?

th0rOdinson (approved):
This looks much professional now, excellent idea!

PR #1234, opened 2025-01-09
feat(migratre): track virtual earn migration

FiboApe:


apps/root/src/services/tierService.ts
```diff
@@ -423,6 +429,26 @@ export default class TierService extends EventsManager<TierServiceData> {
         this.achievements = allAchievements;
         break;
       // ADD HANDLER FOR MIGRATE
```
th0rOdinson, line 431: This comment is solved right?

th0rOdinson (approved):

PR #1251, opened 2025-01-13
feat(earn): always use WETH as ETH

FiboApe:


apps/root/src/services/earnService.ts
```diff
@@ -586,15 +596,126 @@ export class EarnService extends EventsManager<EarnServiceData> {
...
+            }
+          : balance.token,
+      })),
+    }));
+
+    return updatedUserStrategy;
+  }
+
   async fetchUserStrategy(strategyId: Parameters<typeof this.sdkService.getUserStrategy>[0]) {
+    // TODO: UPDATE HERE
```
th0rOdinson, line 710: There are 3 TODOs left in this file

th0rOdinson (approved):

PR #1260, opened 2025-01-13
fix(earn): virtual events again

th0rOdinson:


apps/root/src/services/earnService.ts
```diff
@@ -1235,6 +1238,17 @@ export class EarnService extends EventsManager<EarnServiceData> {
     this.userStrategies = userStrategies;
   }
 
+  handleStoredTransaction(transaction: TransactionDetails) {
+    if (!isEarnType(transaction)) return;
+    const existingUserStrategy = this.userStrategies.find((s) => s.id === transaction.typeData.positionId);
+    const isValidCreateTransaction = !existingUserStrategy && transaction.type === TransactionTypes.earnCreate;
+    const isValidTransaction =
+      existingUserStrategy && existingUserStrategy.lastUpdatedAtFromApi < transaction.addedTime;
+    if (!isValidCreateTransaction && !isValidTransaction) return;
```
FiboApe, line 1247: if this is not a valid create transaction but it is a valid transaction it will still go through right?

apps/root/src/services/earnService.spec.ts
```diff
@@ -1109,6 +1111,122 @@ describe('Earn Service', () => {
...
+        expect(spyHandleTransaction).not.toHaveBeenCalled();
+      });
+    });
+
+    describe('when the position exists', () => {
+      beforeEach(() => {
+        earnService.userStrategies = [createEarnPositionMock({ id: '10-0xvault-10', lastUpdatedAtFromApi: now })];
+      });
+      describe('when the transaction is a create earn transaction', () => {
+        it('should not handle the transaction', () => {
```
FiboApe, line 1174: this is missing the added time

PR #1264, opened 2025-01-14
Add poll user to eligibility

FiboApe:


apps/root/src/services/earnService.ts
```diff
@@ -715,6 +715,10 @@ export class EarnService extends EventsManager<EarnServiceData> {
 
     const userStrategy = await this.sdkService.getUserStrategy(strategyId);
 
+    if (!userStrategy) {
```
th0rOdinson, line 718: NICE

th0rOdinson (approved):
oh yeah

PR #1278, opened 2025-01-20
fix(navigation): reset all forms when navigating away

th0rOdinson:


apps/root/src/frame/components/navigation/index.tsx
```diff
@@ -151,6 +155,22 @@ const Navigation = ({ children }: React.PropsWithChildren) => {
...
+      dispatch(resetAggregatorForm());
+    }
+    if (prevRoute === TRANSFER_ROUTE.key && currentRoute !== TRANSFER_ROUTE.key) {
+      dispatch(resetTransferForm());
+    }
+    if (prevRoute === DCA_ROUTE.key && currentRoute !== DCA_ROUTE.key) {
+      dispatch(resetDcaForm());
+    }
+    if (prevRoute === EARN_PORTFOLIO.key && currentRoute !== EARN_PORTFOLIO.key) {
+      dispatch(resetEarnForm());
```
FiboApe, line 170: this should be fullyResetEarnForm

apps/root/src/state/tabs/hooks.ts
```diff
@@ -1,8 +1,16 @@
 import { useAppSelector } from '@state/hooks';
 import { RootState } from '../index';
+import { shallowEqual } from 'react-redux';
 
 export function useCurrentRoute() {
```
FiboApe, line 5: this would just be useRoute

apps/root/src/frame/components/navigation/index.tsx
```diff
@@ -151,6 +155,22 @@ const Navigation = ({ children }: React.PropsWithChildren) => {
...
+    }
+  }, [currentRoute]);
```
FiboApe, line 172: shouldn't this also change with prevRoute?

th0rOdinson: I couldn't find a current case that needed it, but it's more flexible if I add it :D

FiboApe (approved):

PR #1282, opened 2025-01-21
fix(labels): memoize correctly

th0rOdinson:


apps/root/src/hooks/useStoredLabels.ts
```diff
@@ -7,7 +8,7 @@ function useStoredLabels() {
 
   const storedLabels = useServiceEvents<LabelServiceData, LabelService, 'getLabels'>(labelService, 'getLabels');
 
-  return storedLabels;
+  return React.useMemo(() => storedLabels, [storedLabels]);
```
FiboApe, line 11: this doesn't change anything, why would it work?

FiboApe (approved):

PR #1285, opened 2025-01-22
fix(earn): wizard with only one imput input + filter by enabled networks

th0rOdinson:


apps/root/src/constants/addresses.ts
```diff
@@ -401,6 +401,8 @@ export const SUPPORTED_NETWORKS_DCA = [
   Chains.BASE.chainId,
 ];
 
+export const SUPPORTED_NETWORKS_EARN = [NETWORKS.base.chainId, NETWORKS.optimism.chainId];
```
FiboApe, line 404: Lets use Chains from the SDK

apps/root/src/hooks/useMergedTokensBalances.ts
```diff
@@ -34,6 +40,9 @@ export default function useMergedTokensBalances(selectedWalletOption?: WalletOpt
     const balanceTokens = Object.values(allBalances).reduce<Record<string, BalanceToken>>(
       (acc, { balancesAndPrices, isLoadingChainPrices }) => {
         Object.entries(balancesAndPrices).forEach(([tokenAddress, tokenInfo]) => {
+          if (supportedNetworks && !supportedNetworks.includes(tokenInfo.token.chainId)) {
```
FiboApe, line 43: Why do we need this?

apps/root/src/pages/home/components/portfolio/index.tsx
```diff
@@ -286,7 +286,7 @@ const Portfolio = ({ selectedWalletOption }: PortfolioProps) => {
   const intl = useIntl();
   const showBalances = useShowBalances();
   const pushToHistory = usePushToHistory();
-  const { mergedBalances, isLoadingAllBalances } = useMergedTokensBalances(selectedWalletOption);
+  const { mergedBalances, isLoadingAllBalances } = useMergedTokensBalances({ selectedWalletOption });
```
FiboApe, line 289: so this was always wrong?

packages/ui-library/src/theme/variants/select-variants.ts
```diff
@@ -37,9 +37,6 @@ export const buildSelectVariant = (mode: 'light' | 'dark'): Components => ({
           '& .MuiMenu-paper .MuiMenu-list': {
             padding: `${SPACING(3)} !important`,
             gap: SPACING(0.5),
-            '& .MuiDivider-root': {
-              margin: `${SPACING(4)} ${SPACING(0)}`,
```
FiboApe, line 41: we sure this doesnt break other dividers?

apps/root/src/pages/home/components/portfolio/index.tsx
th0rOdinson: No! Why? I modified the hook so it needs an object now

packages/ui-library/src/theme/variants/select-variants.ts
th0rOdinson: 99.9999%, i checked all select, and we no longer have dividers (we used to have it below the search bar for token and network selectors, now it's gone)

apps/root/src/hooks/useMergedTokensBalances.ts
th0rOdinson: Because we were showing token balances here from chains that you couldn't invest in. 
For instance, if you have 1000 USDC in Polygon, we will show you 1000 USDC and suggest to invest in USDC-Base. Then you enter, you can't invest. Sad.
I think this is aligned with the new label of 'we only show chains we currently support'

FiboApe (approved):

PR #1288, opened 2025-01-23
fix(earn): better sorting and styles on referrals table

th0rOdinson:


apps/root/src/pages/tier-view/referrals/table/index.tsx
```diff
@@ -276,8 +304,8 @@ const ReferralsTable = () => {
                 <ReferralsTableHeader />
               </TableHead>
               <TableBody>
-                {inviteCodes.map((inviteCode) => (
-                  <TableRow key={inviteCode.code}>
+                {sortedInviteCodes.map((inviteCode) => (
+                  <TableRow key={inviteCode.code} sx={({ spacing }) => ({ height: spacing(13) })}>
```
FiboApe, line 308: why are we setting a height for this?

th0rOdinson: when no 'pending' or 'completed' card is displayed, the row tends to be lower in height. This ensures all are equal

FiboApe (approved):

PR #1290, opened 2025-01-23
fix(earn): handle weth as eth in transaction typedata

th0rOdinson:


FiboApe (approved):

apps/root/src/common/utils/transaction-history/parsing.tsx
```diff
@@ -410,9 +414,12 @@ const parseEarnDepositApiEvent: ParseFunction<BaseApiEvent & EarnDepositApiEvent
   earnBaseEventData,
 }) => {
   const earnDepositedTokenId = getTokenListId({ tokenAddress: event.data.depositToken, chainId: event.tx.chainId });
-  const earnDepositedToken = tokenList[earnDepositedTokenId];
+  const earnDepositedToken = parseWrappedProtocolTokenToProtocolToken(tokenList[earnDepositedTokenId]);
```
FiboApe, line 417: token can be undefined here

apps/root/src/common/utils/transaction-history/parsing.tsx
```diff
@@ -521,8 +531,15 @@ const parseEarnWithdrawApiEvent: ParseFunction<BaseApiEvent & EarnWithdrawApiEve
...
         .map((withdrawnToken) => {
-          const tokenId = getTokenListId({ tokenAddress: withdrawnToken.token, chainId: event.tx.chainId });
-          const token = { ...tokenList[tokenId], price: withdrawnToken.price };
+          const tokenId = getTokenListId({
+            tokenAddress: withdrawnToken.token,
+            chainId: event.tx.chainId,
+          });
+
+          const token = parseWrappedProtocolTokenToProtocolToken({
+            ...tokenList[tokenId],
```
FiboApe, line 540: same here

apps/root/src/common/utils/transaction-history/parsing.tsx
```diff
@@ -521,8 +531,15 @@ const parseEarnWithdrawApiEvent: ParseFunction<BaseApiEvent & EarnWithdrawApiEve
...
+            price: withdrawnToken.price,
```
FiboApe, line 541: Price is gonna get added in parseWrappedProtocolTokenToProtocolToken

apps/root/src/common/utils/transaction-history/parsing.tsx
th0rOdinson: In this case, it's checked before at the beginning of the function

apps/root/src/common/utils/transaction-history/parsing.tsx
th0rOdinson: This price comes from the txData, it's not in the token list, that's why I've added it separately

FiboApe (approved):

PR #1298, opened 2025-01-24
fix(earn): include all positions in all timestamps

th0rOdinson:


apps/root/src/pages/earn/portfolio/components/tvl-graph/index.tsx
```diff
@@ -158,7 +158,7 @@ const EarnPositionTvlGraph = ({
                   tickMargin={4}
                   width={30}
                   strokeWidth="0px"
-                  domain={['auto', 'auto']}
+                  domain={[0, 'auto']}
```
FiboApe, line 161: this should not be 0

apps/root/src/services/earnService.ts
```diff
@@ -467,6 +466,15 @@ export class EarnService extends EventsManager<EarnServiceData> {
         'desc'
       );
 
+      const sortedHistoricalBalances = orderBy(userStrategy.historicalBalances || [], 'timestamp', 'asc');
+
+      // This are not 'virtual' events, because they can be in the api, but not applied to the historical balances (24 hs snapshot)
+      const updatedHistoricalBalances = this.applyNewerEventsToHistoricalBalances(
```
FiboApe, line 472: you are not applying the updatedUserStrategies[userStrategyIndex].historicalBalances in this one

th0rOdinson: Historical balances may change due to handleTransaction, so I didn't want to decide whether we needed to apply a transaction from history or not. We could prevent this by comparing timestamps, but I think that would be more prone to error propagation. By using the historical balances from the API as a starting point, we only apply virtualized history and always ensure consistency

apps/root/src/services/earnService.ts
```diff
@@ -631,7 +629,10 @@ export class EarnService extends EventsManager<EarnServiceData> {
       const asset = toToken({ ...strategy.farm.asset, chainId: strategy.farm.chainId });
 
       // 1. Picking the last historical balance (array is sorted by timestamp descending)
```
FiboApe, line 631: This is not correct anymore right?

th0rOdinson: If you mean the if below, you are right!

FiboApe (approved):

PR #1303, opened 2025-01-28
feat(wallet-selector): add isAuth icon to wallet selector

FiboApe:


0xged (approved):

apps/root/src/common/components/wallet-selector/useWalletSelectorState.tsx
```diff
@@ -360,6 +363,22 @@ const useWalletSelectorState = ({ options, showWalletCounter }: WalletSelectorBa
...
+                        })
+                      )}
+                      placement="top"
+                      arrow
+                    >
+                      <ContainerBox>
+                        <KeyIcon />
+                      </ContainerBox>
+                    </Tooltip>
+                  )}
```
th0rOdinson, line 381: What about when wallet is selected?

FiboApe: No need for it there

PR #1346, opened 2025-02-24
fix(earn): hide rewards with showBalances + fix total asset currentPr…

th0rOdinson:
…ofitUsd in table

### DIFF
--- a/apps/root/src/common/utils/earn/parsing.tsx
+++ b/apps/root/src/common/utils/earn/parsing.tsx
@@ -389,6 +389,9 @@ export function parseUserStrategiesFinancialData(userPositions: EarnPosition[] =
     return acc + Number(amount.amountInUSD) || 0;
   }, 0);
 
+  let currentProfitAssetUsd = 0;
+  let currentProfitTotalUsd = 0;
+
   const currentProfit = userPositions.reduce<Record<Address, AmountsOfToken>>((acc, position) => {
     position.balances.forEach((tokenBalance) => {
       if (!acc[tokenBalance.token.address]) {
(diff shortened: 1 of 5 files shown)


apps/root/src/common/utils/earn/parsing.tsx
```diff
@@ -413,17 +424,6 @@ export function parseUserStrategiesFinancialData(userPositions: EarnPosition[] =
     return acc;
   }, {});
 
-  const currentProfitAssetUsd = userPositions.reduce((acc, position) => {
-    const assetAddress = position.strategy.asset.address;
-    const assetProfit = currentProfit[assetAddress]?.amountInUSD || 0;
-    return acc + Number(assetProfit);
```
th0rOdinson, line 419: When there were two 'eth' positions, we had one value for the total ETH, but it was being summed twice—once for each position. :D

apps/root/src/pages/earn/components/strategies-table/components/total-footer/index.tsx
```diff
@@ -77,14 +89,22 @@ const TotalFooter = <T extends StrategiesTableVariants>({
...
-                <Typography variant="bodyBold" color={isEmptyPortfolio ? 'text.disabled' : 'success.dark'}>
-                  {usdFormatter(totalInvested.currentProfitUsd.asset)}
-                </Typography>
+                showBalances ? (
+                  <Typography variant="bodyBold" color={isEmptyPortfolio ? 'text.disabled' : 'success.dark'}>
+                    {usdFormatter(totalInvested.currentProfitUsd.asset)}
+                  </Typography>
+                ) : (
+                  <HiddenNumber size="small" />
+                )
```
th0rOdinson, line 107: I've experimented with using `NetWorthNumber` here, but the USD formatting is specific in this case—it needs to be parsed by the `usdFormatter` to display values like "10k".

FiboApe (approved):

PR #1347, opened 2025-02-24
fix(*): rainbow kit optimization

th0rOdinson:
### Context
RainbowKit uses a Wagmi hook called `useEns`, which is built into the `RainbowKitProvider`. Wagmi hooks usually rely on the `@tanstack/react-query` package to refetch and cache data. Even though I tried, I couldn’t stop the `useEns` call from happening every time a user connects their wallet to the page. However, since we don’t use Wagmi hooks at all, applying some minimization of TanStack usage (reducing retries and refetches) allowed us to achieve significant performance improvements.

Requests before:
Screenshot 2025-02-24 at 4 44 57 PM
Request after:
Screenshot 2025-02-24 at 4 47 43 PM

Useful docs:
- How to config tanstack [TanStack Query Documentation](https://tanstack.com/query/latest/docs/framework/react/reference/useQuery#usequery)
- Wagmi hook in Rainbow: [RainbowKit GitHub Source](https://github.com/rainbow-me/rainbowkit/blob/ed13be55111b6d16c229ff2e7e615c4e225cf29e/packages/rainbowkit/src/components/AccountModal/AccountModal.tsx#L15)

### DIFF
--- a/apps/root/src/frame/index.tsx
+++ b/apps/root/src/frame/index.tsx
@@ -55,7 +55,17 @@ const StyledGridContainer = styled(Grid)<{ isSmall?: boolean }>`
   `}
 `;
 
-const queryClient = new QueryClient();
+const queryClient = new QueryClient({
+  defaultOptions: {
+    queries: {
+      enabled: false,
+      refetchOnWindowFocus: false,
+      retryOnMount: false,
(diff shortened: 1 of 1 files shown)


FiboApe (approved):
Nice catch!

PR #1358, opened 2025-03-17
feat(swap): admit USD amounts in inputs

th0rOdinson:


packages/ui-library/src/components/token-picker-amount-usd-input/index.tsx
```diff
@@ -0,0 +1,340 @@
...
+              nextValue: evt.target.value,
+              decimals: 2,
+            })
+          }
+          value={value || ''}
+          onFocus={onFocus}
+          onBlur={onBlur}
+          endAdornment={
+            <Typography variant="h2Bold" color={getInputColor({ disabled, mode, hasValue: !isUndefined(value) })}>
+              USD
```
FiboApe, line 139: End adornment should be $

FiboApe: Actually it should be the startAdornment

packages/ui-library/src/components/token-picker-amount-usd-input/index.tsx
```diff
@@ -0,0 +1,340 @@
...
+                onBlur={() => setIsFocused(false)}
+                disabled={disabled}
+                priceImpactLabel={priceImpactLabel}
+                onChangeType={onChangeType}
+              />
+            )}
+          </ContainerBox>
+        </ContainerBox>
+        {maxBalanceBtn && !isUndefined(balance) && token && (
+          <StyledMaxButtonContainer>
```
FiboApe, line 318: Max button placement is off

packages/ui-library/src/components/token-picker-amount-usd-input/index.tsx
```diff
@@ -0,0 +1,340 @@
...
+
+const StyledInputContainer = styled(InputContainer)`
+  ${({
+    theme: {
+      spacing,
+      palette: { mode },
+    },
+  }) => `
+    padding: ${spacing(6)};
+    border: 1px solid ${colors[mode].border.border1};
```
FiboApe, line 179: border should be 1.25 px

FiboApe: Also we are missing focused, disabled and error states

FiboApe (commented):
Screenshot 2025-03-18 at 7 18 03 pm
0 on the botton should be disabled too right?

apps/root/src/pages/aggregator/swap-container/components/step1/index.tsx
```diff
@@ -87,7 +85,8 @@ const SwapFirstStep = ({
   const dispatch = useAppDispatch();
   const { trackEvent } = useAnalytics();
   const [transactionWillFail, setTransactionWillFail] = React.useState(false);
-  const prices = usePortfolioPrices(compact([from, to]));
+  const [fromPrice] = useRawUsdPrice(from);
```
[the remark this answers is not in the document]
th0rOdinson: I've removed it because `to` may not have balance, so no price would be found. We can use different approaches for both `from` and `to`

packages/ui-library/src/components/token-amount-usd-input/index.tsx
```diff
@@ -59,19 +59,24 @@ const TokenInput = ({ onChange, value, token, tokenPrice, onBlur, onFocus, disab
     palette: { mode },
   } = useTheme();
   const usdAmount = calculateUsdAmount({ value, token, tokenPrice });
+  const inputRef = useRef<HTMLInputElement>(null);
 
+  const handleChange = (evt: React.ChangeEvent<HTMLInputElement>) => {
```
FiboApe, line 64: this should be memoed right?

packages/ui-library/src/components/token-amount-usd-input/index.tsx
```diff
@@ -101,19 +106,24 @@ const UsdInput = ({ onChange, value, token, tokenPrice, onBlur, onFocus, disable
     palette: { mode },
   } = useTheme();
   const tokenAmount = calculateTokenAmount({ value, tokenPrice });
+  const inputRef = useRef<HTMLInputElement>(null);
+
+  const handleChange = (evt: React.ChangeEvent<HTMLInputElement>) => {
```
FiboApe, line 111: same question

packages/ui-library/src/components/token-picker-amount-usd-input/index.tsx
```diff
@@ -55,19 +55,25 @@ const TokenInput = ({
     palette: { mode },
   } = useTheme();
   const usdAmount = calculateUsdAmount({ value, token, tokenPrice });
+  const inputRef = useRef<HTMLInputElement>(null);
+
+  const handleChange = (evt: React.ChangeEvent<HTMLInputElement>) => {
```
FiboApe, line 60: same

FiboApe (approved):

PR #1372, opened 2025-04-30
feat(earn): enable sonic + gnosis networks

th0rOdinson:


FiboApe (approved):

apps/root/src/constants/earn.ts
```diff
@@ -21,14 +21,6 @@ export const FEE_TYPE_STRING_MAP: Record<FeeType, ReturnType<typeof defineMessag
   }),
 };
 
-// export const EARN_AVAILABLE_CHAINS = [
```
FiboApe, line 24: This should stay so we can control it without the be showing it by default

FiboApe (dismissed):

FiboApe (approved):

